PR #1, opened 2019-09-05
Initial prototype implementation of OAuth2 authentication

mstruk:


oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Base64;
+
+import static io.strimzi.kafka.oauth.common.TokenIntrospection.introspectAccessToken;
+
+public class OAuthAuthenticator {
+
+    private static final Logger log = LoggerFactory.getLogger(OAuthAuthenticator.class);
+
+    public static TokenInfo loginWithAccessToken(String token) {
+        log.warn("loginWithAccessToken() - pass-through access_token: " + token);
```
tombentley: Use `{}` placeholders instead of string concatenation.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/NoopAuthorizer.java
```diff
@@ -0,0 +1,68 @@
...
+import kafka.security.auth.Authorizer;
+import kafka.security.auth.Operation;
+import kafka.security.auth.Resource;
+import org.apache.kafka.common.security.auth.KafkaPrincipal;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scala.collection.immutable.Map;
+import scala.collection.immutable.Set;
+
+public class NoopAuthorizer implements Authorizer {
```
tombentley: Why do we need this class, we don't currently support authz?

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthJWTAuthorizer.java
```diff
@@ -0,0 +1,68 @@
...
+import kafka.security.auth.Authorizer;
+import kafka.security.auth.Operation;
+import kafka.security.auth.Resource;
+import org.apache.kafka.common.security.auth.KafkaPrincipal;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import scala.collection.immutable.Map;
+import scala.collection.immutable.Set;
+
+public class OAuthJWTAuthorizer implements Authorizer {
```
tombentley: I think since authz is not support it's best not to include this class in the PR.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/ServerConfig.java
```diff
@@ -0,0 +1,21 @@
...
+    public static final String OAUTH_JWKS_REFRESH_SECONDS = "oauth.jwks.refresh.seconds";
+    public static final String OAUTH_VALID_ISSUER_URI = "oauth.valid.issuer.uri";
+    public static final String OAUTH_VALIDATE_COMMON_CHECKS = "oauth.validate.common.checks";
+    public static final String OAUTH_VALIDATE_AUDIENCE = "oauth.validate.audience";
+    public static final String OAUTH_INTROSPECTION_ENDPOINT_URI = "oauth.introspection.endpoint.uri";
+
+    public ServerConfig() {
+    }
+
+    public ServerConfig(java.util.Properties p) {
```
tombentley: `import`?

README.md
```diff
@@ -1,3 +1,191 @@
...
+limited set of permissions to a specific client app. In the simplest case, the client application can authenticate in its own name using client 
+credentials. While the client secret is in this case packaged with application client, the benefit is still that it is not shared with application 
+server (Kafka Broker in our case) - the client first performs authentication against OAuth2 authorization server in exchange for an access token, 
+which it then sends to authorization server instead of its secrets. Access tokens can be independently tracked and revoked at will, and represent a 
+limited access to resources on application server.
+
+When user authenticates and authorizes application client in exchange for a token, the client application can be packaged with only the access token
+or the refresh token. User's username and password are never packaged with application client. The access token is sent directly to Kafka Broker 
+during session initialisation, while a refresh token is first used to ask authorisation server for a new access token, before sending it to Kafka 
+Broker to start a new authenticated session.
```
tombentley: I think a diagram showing this interaction would really help people understand. I'm thinking of a directed graph with the server, client and authorization server as nodes and with edges indicating a message. Edges annotated with a number, which corresponds to a description in the text.

README.md
```diff
@@ -1,3 +1,191 @@
...
+Kafka comes with basic OAuth2 support in the form of SASL based authentication module which provides client-server retrieval, exchange and validation 
+of access token as credentials. For real world usage, extensions have to be provided in the form of JAAS callback handlers which is what Strimzi Kafka 
+OAuth does.
+
+
+Configuring the Kafka Broker as a Server
+----------------------------------------
+- Enable OAUTHBEARER as SASL mechanism
+- Configure `io.strimzi.kafka.oauth.server.JaasServerOauthValidatorCallbackHandler` as validator callback handler
+- Specify `admin.user` - in order to properly integrate with default ACL authorization implementation, the broker authenticates itself as _admin_ user.
```
tombentley: Specify it where? In broker config, presumably?

README.md
```diff
@@ -1,3 +1,191 @@
...
+With OAuth2 every access token is associated with some user via unique user identifier. `admin.user` should be set to the principal kafka-broker 
+authenticates as. 
+
+Specify valid issuer uri (OAUTH_VALID_ISSUER_URI) - only access tokens issued by this issuer will be accepted.
```
tombentley: Again, where is this `OAUTH_VALID_ISSUER_URI` supposed to be specified? It's in caps, so I'm not guessing it's an env var?

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
```diff
@@ -0,0 +1,162 @@
...
+        usernameClaim = config.getValue(Config.OAUTH_USERNAME_CLAIM, "sub");
+        if ("sub".equals(usernameClaim)) {
+            usernameClaim = null;
+        }
+
+        if (log.isDebugEnabled()) {
+            log.debug("Configured JaasClientOauthLoginCallbackHandler:\n    token: " + token
+                    + "\n    refreshToken: " + refreshToken + "\n    tokenEndpointUri: " + tokenEndpoint
+                    + "\n    clientId: " + clientId + "\n    clientSecret: " + clientSecret
+                    + "\n    usernameClaim: " + usernameClaim);
```
ppatierno: placeholders insted of string concatanation?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
```diff
@@ -0,0 +1,135 @@
+package io.strimzi.kafka.oauth.common;
+
+import java.util.Properties;
+
+public class Config {
```
ppatierno: does it make sense to name it more specifically as `OAuthConfig` instead of just `Config`?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
```diff
@@ -0,0 +1,135 @@
...
+
+    public static final String OAUTH_CLIENT_ID = "oauth.client.id";
```
scholzj: Is `OAUTH_CLIENT_ID` and `OAUTH_CLIENT_SECRET` ever used in server? If not, should it be in some ClientConfig class in the client?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
scholzj: Should we print the token like this? Shouldn't we just print the information abotu loggign with token without the actual token?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
```diff
@@ -0,0 +1,106 @@
...
+
+        String authorization = "Basic " + base64encode(clientId + ':' + clientSecret);
+
+        StringBuilder body = new StringBuilder("grant_type=client_credentials");
+
+        return post(tokenEndpointUrl, authorization, body.toString());
+    }
+
+    public static TokenInfo loginWithRefreshToken(URI tokenEndpointUrl, String refreshToken, String clientId, String clientSecret) throws IOException {
+        log.warn("loginWithRefreshToken() - tokenEndpointUrl: " + tokenEndpointUrl + ", refreshToken: " + refreshToken
```
scholzj: The client secret can be used to get the tokens so I do notj think it should be printed in the logs.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
```diff
@@ -0,0 +1,199 @@
...
+
+        String jwksUri = config.getValue(ServerConfig.OAUTH_JWKS_ENDPOINT_URI);
+        if (jwksUri != null) {
+
+            validator = new JWTSignatureValidator(
+                    config.getValue(ServerConfig.OAUTH_JWKS_ENDPOINT_URI),
+                    config.getValue(ServerConfig.OAUTH_VALID_ISSUER_URI),
+                    config.getValueAsInt(ServerConfig.OAUTH_JWKS_EXPIRY_SECONDS, 300),
+                    config.getValueAsInt(ServerConfig.OAUTH_JWKS_REFRESH_SECONDS, 360),
+                    config.getValueAsBoolean(ServerConfig.OAUTH_VALIDATE_COMMON_CHECKS, true),
```
scholzj: Do we support this value? If not, we should remove it from here.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/ServerConfig.java
```diff
@@ -0,0 +1,21 @@
...
+
+import io.strimzi.kafka.oauth.common.Config;
+
+public class ServerConfig extends Config {
+
+    public static final String OAUTH_JWKS_ENDPOINT_URI = "oauth.jwks.endpoint.uri";
+    public static final String OAUTH_JWKS_EXPIRY_SECONDS = "oauth.jwks.expiry.seconds";
+    public static final String OAUTH_JWKS_REFRESH_SECONDS = "oauth.jwks.refresh.seconds";
+    public static final String OAUTH_VALID_ISSUER_URI = "oauth.valid.issuer.uri";
+    public static final String OAUTH_VALIDATE_COMMON_CHECKS = "oauth.validate.common.checks";
```
scholzj: We should remove the unsupported fields.

pom.xml
```diff
@@ -0,0 +1,116 @@
...
+    <artifactId>strimzi-kafka-oauth-parent</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+    <packaging>pom</packaging>
+
+    <properties>
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
+        <plugins.compiler.version>3.8.1</plugins.compiler.version>
+
+        <kubernetes-client.version>4.3.0</kubernetes-client.version>
```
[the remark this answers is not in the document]
tombentley: We should use the maven-dependency-plugin to detect declared but unused dependencies.

README.md
```diff
@@ -1,3 +1,191 @@
...
+
+
+Strimzi Kafka OAuth modules provide support for using OAuth2 as authentication mechanism when establishing a session with Kafka Broker.
+
+
+OAuth2 for Authentication
+-------------------------
+
+One of the advantages of OAuth2 compared to direct client-server authentication is that client credentials are never shared with the server. Rather, 
+there exists a separate OAuth2 authorization server that application clients and application servers communicate with. This way user management,
```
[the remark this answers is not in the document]
mstruk: OAuth 2.0 spec strictly uses the term _authorization server_, and mostly does OpenId Connect Spec as well. That's why I use _authorization server_ as well. I guess it more broadly captures that it's not a simple login that the server performs but also provides workflows for users to authorize certain applications with certain level of access.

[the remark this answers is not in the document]
scholzj: Ok ... if it is used by the specs then it is of course fine with me as well ;-)

README.md
mstruk: This is a very general, well specified behaviour covered by OAuth2 spec and widely covered on the Web. Such a diagram should be part of Keycloak project docs or we find one appropriate somewhere else, and link to it?

README.md
```diff
@@ -1,3 +1,191 @@
...
+Since OAuth2 access tokens can contain arbitrary claims - including sets of roles or permissions, rather than basing authorization decisions purely 
+on user's identity, it could be based on externally assigned roles. All we need is define a mapping between access token claims and actions performed 
+on resources. For example, a 'kafka-topic:clicks_*:consumer' claim can be interpreted to mean that the user can read from any topic of which name 
+starts with 'clicks_'.
+
+Another possibility is to delegate authorization decisions to external authorization service - this can be a service provided by the same 
+OAuth2 authorization server used for authentication or it can be a different one, sharing the users database. 
+See https://docs.kantarainitiative.org/uma/wg/oauth-uma-grant-2.0-09.html for OAuth2 based standard for standardised UMA authorization services.
+
+At this time Strimzi Kafka OAuth doesn't provide authorization.
```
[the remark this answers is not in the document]
mstruk: Ok, I can keep it in another branch until the simple authz functionality is available.

README.md
tombentley: I'm open to whatever is easiest (so long as it's compatible with the license of course).

README.md
mstruk: Ok, this whole block of readme needs to be improved to the point one can follow it step by step and get things up and running.

README.md
```diff
@@ -1,3 +1,191 @@
...
+With OAuth2 every access token is associated with some user via unique user identifier. `admin.user` should be set to the principal kafka-broker 
+authenticates as. 
+
+Specify valid issuer uri (OAUTH_VALID_ISSUER_URI) - only access tokens issued by this issuer will be accepted.
+
+Next, if your authorization server issues JWT access tokens, and provides a JWKS certificates endpoint, you can configure that for fast local token 
+validation (OAUTH_JWKS_ENDPOINT_URI). You can control how often JWKS certificates are refreshed by setting OAUTH_JWKS_REFRESH_SECONDS, and you can set
+the maximum time JWKS certificates are considered valid (OAUTH_JWKS_EXPIRY_SECONDS).
+
+TODO: Alternatively, you can also manually retrieve the public key from your authorization server, and configure it explicitly (OAUTH_JWKS_CERT).
```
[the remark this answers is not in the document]
mstruk: Maybe let's just remove it for now. It seemed like something one might easily implement, but it may not be entirely trivial working out storage format for the key and loading / translation to same classes used with JWKS endpoint. At this point it's low priority - I imagine it would only be needed for authorization servers that don't provide JWKS endpoint but support signed access tokens. Might be that there are no such server out there?

README.md
```diff
@@ -1,3 +1,191 @@
...
+  - Optionally specify OAUTH_CLIENT_SECRET depending on whether the client used is a public client or confidential client
+
+
+
+Obtaining the tokens
+--------------------
+
+This step is specific to the authorization server that you use. 
+
+#### Keycloak
```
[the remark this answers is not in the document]
mstruk: I think it would be good to get the demo running with at least one other OAuth2 authorization server.

examples/consumer/src/main/java/io/strimzi/examples/consumer/ExampleConsumer.java
```diff
@@ -0,0 +1,94 @@
...
+        final String REALM = config.getValue("realm", "demo");
+
+        final String TOKEN_ENDPOINT_URI = "http://" + KEYCLOAK_IP+ ":8080/auth/realms/" + REALM + "/protocol/openid-connect/token";
+
+        // TODO: You can set system property 'access.token' or ENV property ACCESS_TOKEN
+        //  and analogously for 'refresh.token' / REFRESH_TOKEN
+        //  See README.md for more info.
+        final String ACCESS_TOKEN = config.getValue("access.token", null);
+        final String REFRESH_TOKEN = config.getValue("refresh.token", null);
+
```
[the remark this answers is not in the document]
mstruk: Do we use some style guide / IDE template / maven stylecheck plugin?

I like empty lines - they generally improve code readability :)

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+
+        usernameClaim = config.getValue(Config.OAUTH_USERNAME_CLAIM, "sub");
+        if ("sub".equals(usernameClaim)) {
+            usernameClaim = null;
+        }
+
+        if (log.isDebugEnabled()) {
+            log.debug("Configured JaasClientOauthLoginCallbackHandler:\n    token: " + token
```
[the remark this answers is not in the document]
mstruk: Yeah, good point. I need to go through logging, switch logging levels to debug or trace, provide log4j config to turn it on. Whole tokens should probably not be logged but for debugging first X chars of the token will make things easier to debug. Also, for debugging stacktraces mean a world of difference, whereas we might not want to log stacktraces relating to failed authentication for example in production config.

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
mstruk: Generally that makes sense in order to skip redundant text concatenations, but I think in this case the code block is well protected behind isDebugEnabled() check and it seems to me that templating strings would actually mean more code, less readable code, and also slightly slower code.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
mstruk: Yes, they are used. Every client connecting to Keycloak server needs to identify with a clientId. That includes client applications that perform authentication with the Keycloak server to get an access token, as well as server applications which accept an access token, and validate it with Keycloak server. So, this is a notion of Keycloak client or more generally of OAuth2 authorization server's client. As opposed to Kafka where Kafka broker is the server, and whatever connects to the Kafka broker is the client. For Keycloak these are both OAuth clients and need to present their clientId and secret.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
mstruk: It makes sense - to make it more IDE friendly (code completion).

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
mstruk: This should be debug or trace level logging, and possibly include only first X chars of the token.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
scholzj: So, does this mean that this needs to be configured in the server as well? If yes, what should be the values? When working on the Strimzi integration I didn't configured this and it didn't seemed that anything complained. (I used the JSON you have in this repo which I imported)

mstruk: When JWTSignatureValidator is used (that is when OAUTH_JWKS_ENDPOINT_URI is configured) only public Keycloak endpoints are used which don't require client to authenticate ... it only fetches certs from server. But if OAuthIntrospectionValidator is used (that is when OAUTH_INTROSPECTION_ENDPOINT_URI is configured) the requests to it have to authenticate with client credentials (clientId + secret).

scholzj: Ok, I have to add it then. Thanks for the explanation.

pom.xml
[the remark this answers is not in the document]
mstruk: Great idea!

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
mstruk: But I tried and I don't like it. It starts to stick out among other classes, and the code is then more verbose. Not worth it.

pom.xml
```diff
@@ -0,0 +1,116 @@
...
+        <kafka.version>2.3.0</kafka.version>
+        <jackson.version>2.9.9.1</jackson.version>
+        <junit.version>4.12</junit.version>
+        <slf4j.version>1.7.26</slf4j.version>
+        <keycloak.version>7.0.0</keycloak.version>
+        <bouncycastle.version>1.60</bouncycastle.version>
```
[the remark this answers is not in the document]
mstruk: It is needed at runtime, keycloak-core depends on it for some of its functionality, so it's  productised already. There's a mention of export policies [here](https://www.bouncycastle.org/specifications.html). If I understand correctly it relies on JVM to enforce encryption security policy (you need to install unrestricted policy files).

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
```diff
@@ -0,0 +1,199 @@
...
+
+        validateConfig();
+
+        String jwksUri = config.getValue(ServerConfig.OAUTH_JWKS_ENDPOINT_URI);
+        if (jwksUri != null) {
+
+            validator = new JWTSignatureValidator(
+                    config.getValue(ServerConfig.OAUTH_JWKS_ENDPOINT_URI),
+                    config.getValue(ServerConfig.OAUTH_VALID_ISSUER_URI),
+                    config.getValueAsInt(ServerConfig.OAUTH_JWKS_EXPIRY_SECONDS, 300),
```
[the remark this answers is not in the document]
mstruk: Good catch.

examples/consumer/src/main/java/io/strimzi/examples/consumer/ExampleConsumer.java
```diff
@@ -0,0 +1,102 @@
...
+    public static void main(String[] args) {
+
+        String topic = "Topic1";
+
+        Properties defaults = new Properties();
+        Config external = new Config();
+
+        // TODO: Set KEYCLOAK_IP to be able to connect to Keycloak
+        //  Use 'keycloak.ip' system property or KEYCLOAK_IP env variable
```
ppatierno: I guess you can remove this TODO, it describes what the code does acutally.

examples/docker/kafka-oauth-strimzi/zookeeper/start.sh
```diff
@@ -0,0 +1,23 @@
...
+    until [ "200" = `curl -sL -o /dev/null -w "%{http_code}" $URL` ]
+    do
+        echo "$MSG ($URL)"
+        sleep 2
+    done
+}
+
+#wait_for_url "http://${KEYCLOAK_IP:-keycloak}:8080/auth" "Waiting for Keycloak to start"
+
+#wait_for_url "http://${KEYCLOAK_IP:-keycloak}:8080/auth/realms/${REALM:-demo}" "Waiting for realm '${REALM}' to be available"
```
ppatierno: can we remove this commented lines? at that point I guess that the `wait_for_url` function is't needed anymore

examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
```diff
@@ -0,0 +1,107 @@
...
+
+    public static void main(String[] args) {
+
+        String topic = "Topic1";
+
+        Properties defaults = new Properties();
+        Config external = new Config();
+
+        // TODO: Set KEYCLOAK_IP to be able to connect to Keycloak
+        //  Use 'keycloak.ip' system property or KEYCLOAK_IP env variable
```
ppatierno: as for the consumer part I guess you can remove this TODO. The code does it now.

examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
```diff
@@ -0,0 +1,107 @@
...
+
+        for (int i = 0; ; i++) {
+            try {
+
+                producer.send(new ProducerRecord<>(topic, "Message " + i))
+                        .get();
+
+                System.out.println("Produced Message " + i);
+
+
```
ppatierno: tbh, I also like blank lines for readability but two blank lines in this case are too much :-)

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
```diff
@@ -0,0 +1,121 @@
...
+        con.setRequestProperty("Accept", "application/json");
+        if (body != null && body.length() > 0) {
+            if (contentType == null) {
+                throw new IllegalArgumentException("contentType must be set when body is not null");
+            }
+            con.setRequestProperty("Content-Type", contentType);
+        }
+
+        try {
+            con.connect();
```
tombentley: I assume you're not calling disconnect because it's reasonable to assume that if OAuth is configured we'll be regularly connecting to the authentication server. That's probably worth a comment.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/IOUtil.java
```diff
@@ -0,0 +1,27 @@
...
+package io.strimzi.kafka.oauth.common;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public class IOUtil {
+
+    public static void copy(InputStream input, OutputStream output) throws IOException {
+        byte [] buf = new byte[4092];
```
tombentley: 4096?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
ppatierno: Tbh I didn't get your point. If I read a class just named `Config` it's not simple to understand that it refers to OAuth without the entire package name.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
```diff
@@ -0,0 +1,121 @@
...
+        } catch (ConnectException e) {
+            throw new IOException("Failed to connect to: " + uri, e);
+        }
+
+        if (body != null && body.length() > 0) {
+            OutputStream out = con.getOutputStream();
+            out.write(body.getBytes(StandardCharsets.UTF_8));
+            out.close();
```
tombentley: Should this be in a `finally`?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public static TokenInfo loginWithAccessToken(String token) {
+        if (log.isDebugEnabled()) {
+            log.debug("loginWithAccessToken() - pass-through access_token: {}", mask(token));
+        }
+        // try introspect token
+        try {
+            return introspectAccessToken(token);
+        } catch (Exception e) {
+            log.debug("[IGNORED] Could not parse token as JWT access token. Could not extract scope, subject, and expiry.", e);
```
ppatierno: maybe it's better a `log.warning` here?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        int code = con.getResponseCode();
+        if (code != 200) {
+            InputStream err = con.getErrorStream();
```
tombentley: This never gets closed.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
```diff
@@ -0,0 +1,117 @@
...
+            log.debug("loginWithRefreshToken() - tokenEndpointUrl: {}, refreshToken: {}, clientId: {}, clientSecret: {}",
+                    tokenEndpointUrl, refreshToken, clientId, mask(clientSecret));
+        }
+
+        String authorization = clientSecret != null ?
+                "Basic " + base64encode(clientId + ':' + clientSecret) :
+                null;
+
+        StringBuilder body = new StringBuilder("grant_type=refresh_token")
+                .append("&refresh_token=").append(refreshToken)
```
tombentley: `refreshToken` doesn't need url encoding?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/SSLUtil.java
```diff
@@ -0,0 +1,78 @@
...
+            throw new RuntimeException("Failed to initialise ssl context", e);
+        }
+
+        return sslContext.getSocketFactory();
+    }
+
+    private static X509TrustManager getTrustManager(TrustManagerFactory tmf) {
+        for (TrustManager tm : tmf.getTrustManagers()) {
+            if (tm instanceof X509TrustManager) {
+                return X509TrustManager.class.cast(tm);
```
tombentley: Since you know `X509TrustManager` statically, you don't need a reflective cast.

```suggestion
                return (X509TrustManager) tm;
```

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/TokenValidator.java
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright 2017-2019, Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.oauth.validator;
+
+import io.strimzi.kafka.oauth.common.TokenInfo;
+
+public interface TokenValidator {
+
+    TokenInfo validate(String token);
```
tombentley: A little javadoc about the contract this is defining (e.g. the exceptions thrown when the token is invalid) probably wouldn't hurt.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
```diff
@@ -0,0 +1,222 @@
...
+        if (jaasConfigEntries.size() != 1) {
+            throw new IllegalArgumentException("Exactly one jaasConfigEntry expected (size: " + jaasConfigEntries.size());
+        }
+
+        for (AppConfigurationEntry e: jaasConfigEntries) {
+            Properties p = new Properties();
+            p.putAll(e.getOptions());
+            config = new ServerConfig(p);
+            break;
+        }
```
tombentley: It's not clear to me how this config works. You're taking some config from the JAAS config, but `ServerConfigs` also looks up in system properties and environment. And you're completely ignoring the `configs` parameter, which on the face of it is how I'd expect this to be configured. Can you explain, or refer me to an existing explanation (maybe I missed it?)

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
```diff
@@ -0,0 +1,222 @@
...
+                    config.getValue(Config.OAUTH_CLIENT_SECRET),
+                    config.getValueAsBoolean(ServerConfig.OAUTH_VALIDATE_COMMON_CHECKS, true),
+                    config.getValue(ServerConfig.OAUTH_VALIDATE_AUDIENCE)
+            );
+        }
+
+        usernameClaim = config.getValue(Config.OAUTH_USERNAME_CLAIM, "sub");
+        if ("sub".equals(usernameClaim)) {
+            usernameClaim = null;
+        }
```
tombentley: Why does "sub" get mapped to null?

README.md
```diff
@@ -1,3 +1,191 @@
...
+
+One of the advantages of OAuth2 compared to direct client-server authentication is that client credentials are never shared with the server. Rather, 
+there exists a separate OAuth2 authorization server that application clients and application servers communicate with. This way user management, 
+authentication, and authorization is 'outsourced' to authorization server.  
+
+User authentication is separated into an outside step which user manually performs to obtain an access token or a refresh token, which grants a 
+limited set of permissions to a specific client app. In the simplest case, the client application can authenticate in its own name using client 
+credentials. While the client secret is in this case packaged with application client, the benefit is still that it is not shared with application 
+server (Kafka Broker in our case) - the client first performs authentication against OAuth2 authorization server in exchange for an access token, 
+which it then sends to authorization server instead of its secrets. Access tokens can be independently tracked and revoked at will, and represent a
```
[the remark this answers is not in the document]
mstruk: Good catch.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
mstruk: Try it, and you'll see. It sticks out from other classes like a sore thumb, and makes the code ugly. As soon as you see it you immediately start to wonder why not prefix other classes in that package as well. But then you're moving into class name a naming scheme that packages are supposed to solve.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
mstruk: The reason I opted for debug logging here is that the token here can be an opaque token (not JWT) which will fail every time on this call. We could introduce another option to tell authenticator what kind of tokens we're dealing with. It's something I heaven's quite worked out how to deal with. The token will in any case be sent to the server, and rejected there if it's not valid.

examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
mstruk: Do we have a style guide that covers empty lines? Otherwise it';s purely a matter of taste. I don't much care if it's one, two, three. If one is more readable to others than two, then sure, there can be one - it's just empty lines.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
mstruk: AFAIK java.net.HttpURLConnection caches connections behind the scenes. connect() call actually takes available connection from internal pool if one exists. And calling close would simply free it and return it to the pool where after some idle time it would be closed.

Also of note - OAuth2 token endpoint is used only during Kafka session initiation, not during production / consumption of records.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
mstruk: Doh! Good catch!

mstruk: In fact this one gets closed in IOUtil.copy method. If reading from inpuststream fails it means inputstream is broken. The implementation behind it should close it itself. If it is consumed fully it should also be closed by underlying implementation itself. But to not rely on proper impl of other people's code the copy() method takes care of closing it.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
mstruk: This one could indeed be left dangling.

examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
tombentley: We don't have a style guide. Personally I wouldn't want one: I would hope we can find a consensus which works for all of us without needing a style guide (which then needs to be enforced if it's to have any value). My personal preference is to use single blank lines sparingly. I find I fit more code on the screen that way. My main desire is to avoid a mixture of styles in a given file (and ideally in a given repo).

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
```diff
@@ -0,0 +1,121 @@
...
+                } catch (Exception e) {
+                    log.warn("[IGNORED] Failed to read response body", e);
+                }
+
+                throw new RuntimeException("Request failed with status " + code + ": " + errbuf.toString(StandardCharsets.UTF_8.name()));
+            } else {
+                throw new RuntimeException("Request failed with status " + code + " " + con.getResponseMessage());
+            }
+        }
+        return JSONUtil.readJSON(con.getInputStream(), responseType);
```
mstruk: This one here is also relying on other people's code being properly implemented. Will protect it ...

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/IOUtil.java
mstruk: Hahaha, so true.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
tombentley:
```
    /**
     * Indicates that other requests to the server
     * are unlikely in the near future. Calling disconnect()
     * should not imply that this HttpURLConnection
     * instance can be reused for other requests.
     */
    public abstract void disconnect();
```

My point is either you're not calling disconnect on purpose (because you're expecting to benefit from the persistent connection to the server while it remains open/in the pool), in which case you should explain why in a comment, or you're not calling disconnect by accident in which case you should fix it.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
mstruk: Well, if it is a JWT refresh token then it won't contain any characters that would require it to be url encoded. But if it is not JWT then I guess it could. So in first case url encoding would be a noop, but in second case it would be required. I'll add it.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/TokenValidator.java
mstruk: Yeah, I need to put more JavaDoc everywhere.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
mstruk: I mostly reverse engineered this by stepping through code. Based on that configs seems to be a master configuration object of Kafka that gives you access to other configuration, possibly allowing you to inject your own things in there. Callback handler module specific configuration is passed via jaasConfigEntries.

ServerConfig constructor takes default values. That means you can override any value through system property or env variable. If not overridden that way use the one provided by jaasConfigEntries.
One TODO is to support per listener configurations - current code assumes there is one OAuth configuration.

Configuration.getValue() is the key to how System properties, env properties override mechanism works.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
mstruk: "sub" is the default claim used for username, meaning that in this case username mapping is not required.
usernameClaim != null means more complex code path further down - to map the username.

README.md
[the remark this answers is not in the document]
mstruk: Instructions removed from main README, moved to example / docker README.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/NoopAuthorizer.java
mstruk: Removed from PR.

mstruk:
> I had a quick review, left a few specific comments and ...
> 
> * the README should be formatted in more a markdown way
I reformatted READMEs with one sentece per line. do you have something else in mind by 'formatted i more a markdown way'?

> * We need Strimzi license headers on all Java source code files
> * I would add more Javadoc on classes and methods
I can add some more. But in principle, no need for JavaDoc where code and structure is obvious. Do we have some convention for marking classes and interfaces that are not supposed to be relied upon for extending on interfacing with ... This whole project should be considered impl detail at this point.

examples/docker/kafka-oauth-strimzi/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+From `docker` directory run:
+
+    docker-compose -f compose.yml -f kafka-oauth-strimzi/compose-ssl.yml up --build
+     
+Kafka broker should be available on localhost:9092. It connects to Keycloak using `https://${KEYCLOAK_IP}:8443`
+
+
+### Explaining the configuration settings
+
```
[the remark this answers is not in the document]
mstruk: Removed. Maybe we can add some additional explanations here, but it should be quite self-explanatory with the help of root README.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
mstruk: Added comments in code and JavaDoc.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
mstruk: To put it succinctly - the OAuth2 config uses its own system properties / environment variables / JAAS login module options. It makes no use of Kafka's server.properties.

mstruk:
> I don't see any tests; do you have plans for adding some?

It would be great to have some integration tests, but it seems quite a project. I would skip it for initial PR, especially since Strimzi integration seems to work. There are probably some bugs in there, but most likely more hidden ones, that would only be uncovered by more sophisticated tests.

scholzj:
I would suggest to leave JavaDocs for a subsequent PR:
* It will be easier to review
* Give how this is expected to be used I think READMEs are more useful then JavaDocs here.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
mstruk: Yeah, I think common checks should always be on. I'll remove option.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/ServerConfig.java
mstruk: Also audience check, not sure what the use case is really.

PR #4, opened 2019-09-16
Rename artifacts, add plugins for source and javadoc jars, checkstyle …

scholzj:
This PR does several cleanup things and preparations for release:
* Javadoc and sources JAR plugins
* Checkstyle
* Renames the artifatcs etc.


scholzj:
(FYI: most of the rules etc. were just copied from Strimzi to have the same here)

PR #9, opened 2019-09-17
Remove unused dependencies

scholzj:
There seem to be some unused depndencies in the `pom.xml`. We should clean them up.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,12 +67,10 @@
         <sonatype.nexus.staging>1.6.3</sonatype.nexus.staging>
 
         <kafka.version>2.3.0</kafka.version>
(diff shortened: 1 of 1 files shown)


tombentley:
We should use `maven-dependency-plugin`'s `analyze-only` goal with `failOnWarning` to catch these a build time.

scholzj:
@tombentley We do have that (https://github.com/strimzi/strimzi-kafka-oauth/blob/efe993b208a008e6d7af0c8f6d610886ac551e33/pom.xml#L181), but it does not seem to be smart enough to catch these. I think it catches these dependencies only if they are declared and unused in some module. But not in the `dependencyManagement` section.

PR #15, opened 2019-10-30
Added 'hydra' authorization server to examples, and addressed additional issues that popped up

mstruk:
- Refactored examples keys, and certificates
- Added OAUTH_TOKENS_NOT_JWT config option
- Added OAUTH_MAX_TOKEN_EXPIRY_SECONDS client config option
- Added OAUTH_VALIDATION_SKIP_TYPE_CHECK server config option


oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
```diff
@@ -66,7 +68,7 @@ public void configure(Map<String, ?> configs, String saslMechanism, List<AppConf
             String endpoint = config.getValue(ClientConfig.OAUTH_TOKEN_ENDPOINT_URI);
 
             if (endpoint == null) {
-                throw new RuntimeException("Access Token not specified ('oauth.access.token'). OAuth2 Token Endpoint ('oauth.token.endpoint.uri') should then be set.");
+                throw new RuntimeException("Access Token not specified (OAUTH_ACCESS_TOKEN). OAuth2 Token Endpoint (OAUTH_TOKEN_ENDPOINT_URI) should then be set.");
```
scholzj: What is the reason for changing these? the configuration options are still valid, or?

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
```diff
@@ -81,7 +83,7 @@ public void configure(Map<String, ?> configs, String saslMechanism, List<AppConf
             clientSecret = config.getValue(Config.OAUTH_CLIENT_SECRET);
 
             if (clientId == null) {
-                throw new RuntimeException("No client id specified ('oauth.client.id')");
+                throw new RuntimeException("No client id specified (OAUTH_CLIENT_ID)");
```
scholzj: Same as above.

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
mstruk: It's for consistency reasons, because in other places it's also env var name that's used rather than system property name. It does not affect logic - options stay the same and are still valid.

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
mstruk: All documentation everywhere only really mentions env vars, and in some other places there already was env var used in error messages. I opted to use env vars everywhere in error messages for consistency, rather than using equivalent system property names.

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
scholzj: I think thsi could be a bit confusing because on the server side we use the options and not env. vars. But you are right - it is that way in most other messages so lets keep it consitent.

examples/consumer/README.md
```diff
@@ -6,49 +6,64 @@ This projects demonstrates a Kafka client that uses OAuth2 to establish Kafka se
...
 --------
+
     mvn clean install
 
 
 Preparing
 ---------
 
-Determine your machine's local network IP address, and add keycloak entry to your `/etc/host`.
+Make sure that the following ports on your host machine are free: 9092, 2128 (Kafka), 8080, 8443 (Keycloak), 4444, 4445 (Hydra).
```
ppatierno: what's this port 2128 about Kafka?

examples/docker/certificates/gen-ca.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# create CA certificate
+openssl req -x509 -new -nodes -sha256 -days 3650 -subj "/CN=strimzi.io" -key ca.key -out ca.crt
+
+
+PASSWORD=changeit
+
+# create p12 truststore
+# can't use openssl, the resulting .p12 truststore not usable by java
+#openssl pkcs12 -export -nokeys -in ca.crt -out ca-truststore.p12 -passout pass:$PASSWORD
```
ppatierno: can we remove this part then?

examples/consumer/README.md
mstruk: Fixed.

examples/docker/certificates/gen-ca.sh
mstruk: Removed.

scholzj:
@ppatierno Is this now ok from your perspective? It seems the comments should be fixed.

PR #19, opened 2019-11-18
Add RELEASE.md describing the release process and use separate version params in independent pom.xml files

scholzj:
This PR adds new `RELEASE.md` file describing the release process. It also adds new properties for managing the versions of the Strimzi Kafka OAuth libraries.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,39 @@
+# Strimzi Kafka OAuth release process
+
+This document describes the steps needed for release of Strimti Kafka OAuth library.
+
+## Create new release branch
+
+The release branch should be named `release-M.N.x` where `M.N` is replaced with the major and minor versions of the new release.
+For example:
+
+```
+git checkout -b release-0.2.x
+```
+
+## Set new version in main Maven project
+
+Update the version for the new Maven project:
+
+```
+mvn versions:set -DnewVersion=X.Y.Z
+```
+
+Where `X.Y.Z` is the new version which is going to be released.
+
+## Update dependencies in examples and tests
+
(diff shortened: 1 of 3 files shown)


RELEASE.md
```diff
@@ -0,0 +1,39 @@
...
+
+```
+mvn versions:set -DnewVersion=X.Y.Z
+```
+
+Where `X.Y.Z` is the new version which is going to be released.
+
+## Update dependencies in examples and tests
+
+Update the version in dependencies in:
```
mstruk: I propose more explicit directions:

Update the `strimzi-oauth.version` property in:

RELEASE.md
```diff
@@ -0,0 +1,39 @@
...
+* `examples/docker/pom.xml`
+* `testsuite/pom.xml`
+
+These are independent Maven projects, so the dependency version needs ot be bumped manually.
```
mstruk: Type: needs ot -> needs to

RELEASE.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Set new version in main Maven project
+
+Update the version for the new Maven project:
+
+```
+mvn versions:set -DnewVersion=X.Y.Z
+```
+
+Where `X.Y.Z` is the new version which is going to be released.
```
ppatierno: should be M=X and Y=N? Why not using them, it could be confusing

scholzj: `X.Y.Z` is the normal placeholder for the version. But in the branch name it conflicets with the `.x` at the end. That is why different letters are used here and there.

PR #21, opened 2019-11-29
Add s390x support for Travis CI

cnnaik:
Add s390x support for Travis CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 dist: xenial
 sudo: required
 language: java
-
 jobs:
   include:
   - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,18 @@
 dist: xenial
 sudo: required
 language: java
+
+jobs:
+  include:
+  - os: linux
+  - os: linux
+    arch: s390x
+    jdk: openjdk11
+    addons:
+      apt:
+        packages:
+          - maven
+          
 jdk:
 - openjdk8
 - openjdk11


scholzj:
A stupid question ... but what is the expected value? Are there some major differences with Java on s390x? How does this impact the tests which seems to be mostly using container images not necessarily built for s390x?

cnnaik:
We have added this support to make sure that strimzi-kafka-oauth builds fine on s390x, also any further enhancements do not break it on s390x.

Also note that the changes are added only for OpenJDK11 on s390x which does not run any test cases.

scholzj:
@cnnaik Actually, I thought this is running for both Java 8 and 11. But now I see that it runs Java 11 for S390x but one another build for Java 8 on amd64 (see here: https://travis-ci.org/strimzi/strimzi-kafka-oauth/builds/618455629). Is there some way to avoid that and run the Java 8 build only once?

cnnaik:
@scholzj As per the build configuration of the jobs only 3 jobs should run for the changes made to the `travis.yml`. We tried to build in org environment (https://travis-ci.org/cnnaik/strimzi-kafka-oauth/builds/620092310) , we got 3 builds only in `strimzi-s390x-travis` branch. 

Could you please check if you can re-trigger the build?

scholzj:
@cnnaik That is weird. Maybe it is related to how the merging of the PR works on Travis or something. Can you fix the DCO and I will try to merge and see what it does as a branch build.

cnnaik:
@scholzj Fixed the DCO check. Please check.

scholzj:
Hmm, seems to still do 4 branches on Strimzi Travis. It is strange that it behaves differently for your branch.

scholzj:
@cnnaik For some reason it seems to work differently. As if you are on a different Travis version (did you maybe enabled the beta version?). But I got it fixed somehow - see #23.

cnnaik:
@scholzj  No i have not enabled anything. 
Will this build create s390x specific jar files which will be available for download for the users?

scholzj:
Do we need it? I don't think the jars contain any native code. So I think
it should not be needed.

PR #24, opened 2020-01-10
Keycloak Authorization Services based authz prototype

mstruk:

This PR introduces authorization support that works hand-in-hand with OAuth2 authentication.

There is an example that allows a step-by-step walkthrough. See [README-authz.md](https://github.com/mstruk/strimzi-kafka-oauth/blob/authz/examples/README-authz.md)

The configuration is explained in JavaDoc of [KeycloakRBACAuthorizer class](https://github.com/mstruk/strimzi-kafka-oauth/blob/authz/oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java)

What's still missing are automated tests in the testsuite.


oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpException.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.net.URI;
+
+public class HttpException extends RuntimeException {
+
+    private URI uri;
+    private int status;
+    private String response;
+
+    public HttpException(URI uri, int status, String response) {
```
tombentley: Do you need to include the request method too?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpException.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package io.strimzi.kafka.oauth.common;
+
+import java.net.URI;
+
+public class HttpException extends RuntimeException {
+
+    private URI uri;
+    private int status;
+    private String response;
```
tombentley: These can be final

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/JSONUtil.java
```diff
@@ -19,23 +22,64 @@
...
     }
 
+    public static List<String> asListOfString(JsonNode arrayNode) {
+        if (!arrayNode.isArray()) {
+            throw new IllegalArgumentException("JsonNode not an array node: " + arrayNode);
+        }
+        ArrayList<String> result = new ArrayList<>();
+        Iterator<JsonNode> it = arrayNode.iterator();
+        while (it.hasNext()) {
+            result.add(it.next().asText());
```
tombentley: IAE for non-array, but not for non-String member of the array?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/JwtKafkaPrincipal.java
```diff
@@ -0,0 +1,26 @@
...
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.oauth.server.authorizer;
+
+import io.strimzi.kafka.oauth.common.BearerTokenWithPayload;
+import org.apache.kafka.common.security.auth.KafkaPrincipal;
+
+public class JwtKafkaPrincipal extends KafkaPrincipal {
+
+    private BearerTokenWithPayload jwt;
```
tombentley:
```suggestion
    private final BearerTokenWithPayload jwt;
```

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/JwtKafkaPrincipal.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public JwtKafkaPrincipal(String principalType, String name) {
+        super(principalType, name);
```
tombentley:
```suggestion
        this(principalType, name, null);
```

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/JwtKafkaPrincipal.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public JwtKafkaPrincipal(String principalType, String name, BearerTokenWithPayload jwt) {
+        this(principalType, name);
```
tombentley:
```suggestion
        super(principalType, name);
```

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
```diff
@@ -0,0 +1,416 @@
...
+
+
+    public KeycloakRBACAuthorizer() {
+        super();
+    }
+
+    @Override
+    public void configure(Map<String, ?> configs) {
+        super.configure(configs);
+
```
tombentley: Since using this authorizer requires using the principal builder I suppose you could check it here and throw if the principal builder is wrong.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
```diff
@@ -0,0 +1,416 @@
...
+            if (log.isDebugEnabled()) {
+                String status = granted ? "GRANTED" : "DENIED";
+                log.debug("Authorization " + status + " by ACL - user: " + session.principal() + ", operation: " + operation + ", resource: " + resource);
+            }
+            return granted;
+        }
+
+        if (log.isDebugEnabled()) {
+            log.debug("Authorization DENIED - cluster: " + clusterName + ", user: " + session.principal() + ", operation: " + operation + ", resource: " + resource + "\n permissions: " + authz);
+        }
```
tombentley: I think we need to be a bit careful with the logging here.

For auditing purposes it's likely that users might be more interested when authorization is denied than when it's granted. Kafka attempted to solve that by logging deny at a higher level. Although that doesn't really work (because `authorize` can also be used for filtering what's included in protocol responses, as mentioned in https://cwiki.apache.org/confluence/display/KAFKA/KIP-504+-+Add+new+Java+Authorizer+Interface#KIP-504-AddnewJavaAuthorizerInterface-AclAuthorizer), I think we need to distinguish deny in a way which users can configure their logging for. i.e. either level, or logger name. But we should should also be aware when we've delegated result will also (probably) have been logged by the delegate.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/UserSpec.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2017-2019, Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.oauth.server.authorizer;
+
+public class UserSpec {
+
+    private String type;
+    private String name;
```
tombentley: can be final

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/UserSpec.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+
+    public static UserSpec of(String principal) {
+        String[] spec = principal.split(":");
+        return new UserSpec(spec[0], spec[1]);
+    }
+
+    public String toString() {
+        return super.toString() + type + ":" + name;
```
tombentley: Dunno if you want to separate the `Object.toString()` from the extra bits?
```suggestion
        return super.toString() + " " + type + ":" + name;
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+Or, thinking about it another way, if you want to limit certain permissions only to regular user accounts where authentication with username and password is required, you should use Group policies, rather than Role policies.
+That's what we see used in `permissions` that start with 'ClusterManager'.
+Performing cluster management is usually done interactively - in person - using CLI tools. 
+It makes sense to require the user to log-in, before using the resulting access token to authenticate to Kafka Broker.
+In this case the access token represents the specific user, rather than the client application.
+
+Before continuing, there is one setting we need to check.
+Due to [a little bug in Keycloak](https://issues.redhat.com/browse/KEYCLOAK-12640) the realm is at this point misconfigured, and we have to fix the configuration manually.
+Under `Clients` / `kafka` / `Authorization` / `Settings` make sure the `Decision Strategy` is set to `Affirmative`, and NOT to `Unanimous`. Click `Save` after fixing it.
+
```
tombentley: New subsection: Testing with clients

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+It makes sense to require the user to log-in, before using the resulting access token to authenticate to Kafka Broker.
+In this case the access token represents the specific user, rather than the client application.
+
+Before continuing, there is one setting we need to check.
+Due to [a little bug in Keycloak](https://issues.redhat.com/browse/KEYCLOAK-12640) the realm is at this point misconfigured, and we have to fix the configuration manually.
+Under `Clients` / `kafka` / `Authorization` / `Settings` make sure the `Decision Strategy` is set to `Affirmative`, and NOT to `Unanimous`. Click `Save` after fixing it.
+  
+With configuration now in place, let's create some topics, use a producer, a consumer, and try to perform some management operations using different user and service accounts. 
+
+First, we'll spin up a new docker container based on Kafka image previously built by `docker-compose` which we'll use to connect to already running Kafka Broker.
```
tombentley:
```suggestion
First, we'll spin up a new docker container based on a Kafka image previously built by `docker-compose` which we'll use to connect to the already running Kafka Broker.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+
+    docker run -ti --rm --name kafka-cli --network docker_default strimzi/example-kafka /bin/sh
+    
+Let's try to produce some messages as client `team-a-client`.
+First, we prepare a configuration file with authentication configuration.
```
tombentley:
```suggestion
First, we prepare a Kafka consumer configuration file with authentication parameters.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+sasl.mechanism=OAUTHBEARER
+sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
+  oauth.client.id="team-a-client" \
+  oauth.client.secret="team-a-client-secret" \
+  oauth.token.endpoint.uri="http://keycloak:8080/auth/realms/kafka-authz/protocol/openid-connect/token" ;
+sasl.login.callback.handler.class=io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler
+EOF
+```
+
+In Keycloak Console you can find which roles are assigned to `team-a-client` service account, by selecting `team-a-client` under `Clients` section,
```
tombentley:
```suggestion
In the Keycloak Console you can find which roles are assigned to the `team-a-client` service account, by selecting `team-a-client` in the `Clients` section,
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+and then opening `Service Account Roles` tab for the client.
```
tombentley:
```suggestion
and then opening the `Service Account Roles` tab for the client.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+You should see `Dev Team A` realm role assigned.
```
tombentley:
```suggestion
You should see the `Dev Team A` realm role assigned. 
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+  oauth.token.endpoint.uri="http://keycloak:8080/auth/realms/kafka-authz/protocol/openid-connect/token" ;
+sasl.login.callback.handler.class=io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler
+EOF
+```
+
+In Keycloak Console you can find which roles are assigned to `team-a-client` service account, by selecting `team-a-client` under `Clients` section,
+and then opening `Service Account Roles` tab for the client.
+You should see `Dev Team A` realm role assigned. 
+
+We can now use this configuration with Kafka CLI tools.
```
tombentley:
```suggestion
We can now use this configuration with Kafka's CLI tools.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+
+```
+bin/kafka-console-producer.sh --broker-list kafka:9092 --topic my-topic --producer.config=$HOME/team-a-client.properties
+First message
+```
+
+When we press `Enter` to push the first message we receive `Not authorized to access topics: [my-topic]` error.
+
+`team-a-client` has a `Dev Team A` role which gives it permissions to do anything on topics that start with 'a_', and only write to topics that start with 'x_'.
+Topic named `my-topic` does not fall in either of these two.
```
tombentley:
```suggestion
The topic named `my-topic` matches neither of those.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+
+Use CTRL-C to exit the CLI application, and let's try to write to topic `a_messages`.
+
+```
+bin/kafka-console-producer.sh --broker-list kafka:9092 --topic a_messages --producer.config ~/team-a-client.properties
+First message
+Second message
+```
+
+We can see some warnings but looking at Kafka container log there is DEBUG level output saying 'Authorization GRANTED'.
```
tombentley:
```suggestion
Although we can see some unrelated warnings, looking at the Kafka container log there is DEBUG level output saying 'Authorization GRANTED'.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+    bin/kafka-console-consumer.sh --bootstrap-server kafka:9092 --topic a_messages --from-beginning --consumer.config ~/team-a-client.properties
+
+This gives us an error like: 'Not authorized to access group: console-consumer-55841'.
+
+The reason is that we have to override the default consumer group name - `Dev Team A` only has access to consumer groups that have names starting with 'a_'.
+Let's set custom consumer group name that starts with 'a_'
+
+    bin/kafka-console-consumer.sh --bootstrap-server kafka:9092 --topic a_messages --from-beginning --consumer.config ~/team-a-client.properties --group a_consumer_group_1
+
+We should now receive all the messages for 'a_messages' topic, after which the client blocks waiting for more to come down the stream.
```
tombentley:
```suggestion
We should now receive all the messages for the 'a_messages' topic, after which the client blocks waiting for more messages.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+    
+We get one topic listed: 'a_messages'.
+
+Let's try and list the consumer groups:
+
+    bin/kafka-consumer-groups.sh --bootstrap-server kafka:9092 --command-config ~/team-a-client.properties --list
+
+Similarly to listing topics, we get one consumer group listed: `a_consumer_group_1`.
+
+There are more CLI administrative tools. For example we can try get default cluster configuration:
```
tombentley:
```suggestion
There are more CLI administrative tools. For example we can try to get the default cluster configuration:
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+sasl.mechanism=OAUTHBEARER
+sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
+  oauth.client.id="team-b-client" \
+  oauth.client.secret="team-b-client-secret" \
+  oauth.token.endpoint.uri="http://keycloak:8080/auth/realms/kafka-authz/protocol/openid-connect/token" ;
+sasl.login.callback.handler.class=io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler
+EOF
+```
+
+If we look at `team-b-client` client configuration, under 'Service Account Roles' we can see that it has `Dev Team B` realm role assigned.
```
tombentley:
```suggestion
If we look at `team-b-client` client configuration in Keycloak, under 'Service Account Roles' we can see that it has the `Dev Team B` realm role assigned.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+Looking in Keycloak Console at `kafka` client's `Authorization` tab where `Permissions` are listed, we can see the permissions that start with 'Dev Team B ...'.
```
tombentley:
```suggestion
Looking in Keycloak Console at the `kafka` client's `Authorization` tab where `Permissions` are listed, we can see the permissions that start with 'Dev Team B ...'.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+These match the users and service accounts that have 'Dev Team B' realm role assigned to them.
```
tombentley:
```suggestion
These match the users and service accounts that have the 'Dev Team B' realm role assigned to them. 
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+We get a `Not authorized to access topics: [x_messages]` error as we expected. 
+Client `team-a-client`, on the other hand, should be able to write to such a topic:
+
+```
+bin/kafka-console-producer.sh --broker-list kafka:9092 --topic x_messages --producer.config ~/team-a-client.properties
+Message 1
+```
+
+However, we again receive `Not authorized to access topics: [x_messages]`. What's going on?
+The reason for failure is that while `team-a-client` can write to `x_messages` topic, it does not have a permission to create one if it does not yet exist.
```
tombentley:
```suggestion
The reason for failure is that while `team-a-client` can write to `x_messages` topic, it does not have a permission to create a topic if it does not yet exist.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+
+We now need a power user that can create a topic with all the proper settings - like the right number of partitions and replicas.
+
+Let's create a configuration for user `bob` who has full ability to manage everything on Kafka cluster `cluster2`.
+
+First, `bob` will authenticate to Keycloak server with username and password and get a refresh token.
```
tombentley:
```suggestion
First, `bob` will authenticate to Keycloak server with his username and password and get a refresh token.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+
+```
+cat > ~/bob.properties << EOF
+security.protocol=SASL_PLAINTEXT
+sasl.mechanism=OAUTHBEARER
+sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
+  oauth.refresh.token="$REFRESH_TOKEN" \
+  oauth.client.id="kafka-cli" \
+  oauth.token.endpoint.uri="http://keycloak:8080/auth/realms/kafka-authz/protocol/openid-connect/token" ;
+sasl.login.callback.handler.class=io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler
```
tombentley: I don't think it matters (i.e. I think JAAS is happy with a single line config), but I'll point out that the `\` escapes are interpreted by the shell and do not become \ escapes for the newlines in the properties file.

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+EOF
+```
+
+Note that we use `kafka-cli` public client for clientId.
```
tombentley:
```suggestion
Note that we use the `kafka-cli` public client for the `oauth.client.id` in the `sasl.jaas.config`.
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+  oauth.refresh.token="$REFRESH_TOKEN" \
+  oauth.client.id="kafka-cli" \
+  oauth.token.endpoint.uri="http://keycloak:8080/auth/realms/kafka-authz/protocol/openid-connect/token" ;
+sasl.login.callback.handler.class=io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler
+EOF
+```
+
+Note that we use `kafka-cli` public client for clientId. 
+Since that is a public client it does not require any secret.
+We can use this, because we authenticate with a token directly (in this case a refresh token is used to request an access token behind the scenes which is then sent to Kafka broker for authentication).
```
tombentley:
```suggestion
We can use this because we authenticate with a token directly (in this case a refresh token is used to request an access token behind the scenes which is then sent to Kafka broker for authentication, and we already did the authentication for the refresh token).
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+
+Let's now try to create `x_messages` topic:
```
tombentley:
```suggestion
Let's now try to create the `x_messages` topic:
```

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+
+If we try the same as `team-b-client` it should fail.
+
+```
+bin/kafka-console-producer.sh --broker-list kafka:9092 --topic x_messages --producer.config ~/team-b-client.properties
+Message 4
+Message 5
+```
+
+But `team-b-client` should be able to consume messages from `x_messages` topic:
```
tombentley:
```suggestion
But `team-b-client` should be able to consume messages from the `x_messages` topic:
```

examples/docker/kafka-oauth-strimzi/compose-authz.yml
```diff
@@ -0,0 +1,78 @@
...
+      # Java Debug
+      #KAFKA_DEBUG: y
+      #DEBUG_SUSPEND_FLAG: y
+      #JAVA_DEBUG_PORT: 5006
+
+      #
+      # KAFKA Configuration
+      #
+      LOG_DIR: /home/kafka/logs
+      #KAFKA_LOG_DIRS: /home/kafka/1
```
tombentley: Keep this?

examples/docker/kafka-oauth-strimzi/kafka/oauth.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/bash
+
+usage() {
+  echo "Usage: $0 [USERNAME] [PASSWORD] [ARGUMENTS] ..."
+  echo
```
tombentley: I think you should add a sentence explaining what this command actually _does_. "Requests a refresh or access token from ...."

examples/docker/kafka-oauth-strimzi/kafka/jwt.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+if [ "$1" == "" ]; then
```
tombentley: Or `--help`?

examples/docker/kafka-oauth-strimzi/kafka/oauth.sh
```diff
@@ -0,0 +1,119 @@
...
+    exit 1
+fi
+
+if [ "$UNAME" != "" ] && [ "$PASS" == "" ]; then
+    >&2 read -s -p "Password: " PASS
+    >&2 echo
+fi
+
+if [ "$UNAME" == "" ] && [ "CLIENT_ID" == "" ]; then
+    echo "Usage: $0 [USERNAME] [PASSWORD] [--insecure] [--access] [--endpoint TOKEN_ENDPOINT_URL] [--client-id CLIENT_ID] [--secret CLIENT_SECRET] [--scopes SCOPES]"
```
tombentley: Say that both USERNAME and --client-id must be given

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+They can not, however, have groups assigned.
+As a consequence, if you want to target permissions to microservices using service accounts, you can't use Group policies, but are forced to use Role policies.
+Or, thinking about it another way, if you want to limit certain permissions only to regular user accounts where authentication with username and password is required, you should use Group policies, rather than Role policies.
+That's what we see used in `permissions` that start with 'ClusterManager'.
+Performing cluster management is usually done interactively - in person - using CLI tools. 
+It makes sense to require the user to log-in, before using the resulting access token to authenticate to Kafka Broker.
+In this case the access token represents the specific user, rather than the client application.
+
+Before continuing, there is one setting we need to check.
+Due to [a little bug in Keycloak](https://issues.redhat.com/browse/KEYCLOAK-12640) the realm is at this point misconfigured, and we have to fix the configuration manually.
```
ppatierno: I think we should track this bug maybe with an issue here, in order to remember to remove this section when the Keycloak bug will be fixed and the workaround not needed anymore.

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+and then opening `Service Account Roles` tab for the client.
+You should see `Dev Team A` realm role assigned. 
+
+We can now use this configuration with Kafka CLI tools.
+
+Make sure the necessary classes are on the classpath:
+
+    export CLASSPATH=/opt/kafka/libs/strimzi/*:$CLASSPATH
+
+Now let's try produce some messages to topic 'my-topic':
```
ppatierno: Starting from here, it seems that text is in "italics"

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/JwtKafkaPrincipalBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.kafka.common.security.authenticator.DefaultKafkaPrincipalBuilder;
+import org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule;
+import org.apache.kafka.common.security.oauthbearer.internals.OAuthBearerSaslServer;
+
+import java.util.Map;
+
+/**
+ * This class needs to be enabled as the PrincipalBuilder on Kafka Broker.
+ *
+ * It ensures that the generated Principal instance
```
ppatierno: This sentence seems to miss the end :-( "ensures" what?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/JSONUtil.java
mstruk: We convert elements to strings and they get validated in the calling `authorize()` method immediately afterwards. Helper method itself is potentially more universally usable - for some other occasion maybe, if it doesn't enforce by itself that elements have to be strings.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
mstruk: Good point. I gave no thought to possible auditing requirements here. Usually it has to be a formatted output that's easy to parse. For now I'll just make separate loggers and use debug level to log. Logging can then be configured separately for GRANT and DENY logging.

examples/docker/kafka-oauth-strimzi/kafka/oauth.sh
mstruk: That's not actually the case - the --client-id only has to be specified if USERNAME has not been, otherwise it default to 'kafka-cli' value. The error here is actually the missing '$' in "CLIENT_ID"

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
...
+    
+Let's start up all the containers with authorization configured, and we'll then perform any manual step, and explain how everything works.
+
+    docker-compose -f compose.yml -f keycloak/compose.yml -f keycloak-import/compose.yml -f kafka-oauth-strimzi/compose-authz.yml up --build
+
+When everything starts up without errors we should have one instance of `keycloak` listening on localhost:8080.
+You can login to admin console by opening `http://localhost:8080/auth/admin` and using `admin` as both username, and a password.
+
+In upper left corner under Keycloak icon you should see `Master` selected as a current realm.
+Moving the mouse pointer over it should reveal two additional realms - `Demo` and `Kafka-authz`.
```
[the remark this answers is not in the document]
mstruk: That's a quirk of Keycloak Admin Console. Here I refer to navigation labels where Keycloak capitalizes realm names to get a label.

examples/README-authz.md
mstruk: Good idea.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpException.java
mstruk: Indeed.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
mstruk: The check has to be performed in authorize() call in any case. But we can add it here as well.

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
+## Token based authorization with Keycloak Authorization Services
+
+Once the Kafka Broker has obtained an access token by using Strimzi Kafka OAuth for authentication, it is possible to use centrally managed authorization rules to enforce access restrictions onto Kafka Clients.
+For this Strimzi Kafka OAuth supports the use of Keycloak Authorization Services.
+
+A custom authorizer has to be configured on the Kafka Broker to take advantage of Authorization Services REST endpoints available on Keycloak, which provide a list of granted permissions on resources for authenticated users.
+The list is fetched once, and enforced locally on Kafka Broker for each user session in order to provide fast authorization decisions.
```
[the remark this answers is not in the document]
mstruk: I wanted to keep the first iteration as simple as possible. I imagine refetching authorization grants would require a background service with a configurable threadpool, and period duration. The authz grants have to be fetched user-by-user so we then also have to be able to iterate over existing sessions to gather the access tokens ... It also depends on the typical lifetime of Kafka session. If these are short enough one can forego reauthorization altogether.

examples/docker/kafka-oauth-strimzi/compose-authz.yml
mstruk: Right, no need.

examples/README-authz.md
```diff
@@ -0,0 +1,403 @@
+## Token Based Authorization with Keycloak Authorization Services
+
+Once the Kafka Broker has obtained an access token by using Strimzi Kafka OAuth for authentication, it is possible to use centrally managed authorization rules to enforce access restrictions onto Kafka Clients.
+For this Strimzi Kafka OAuth supports the use of Keycloak Authorization Services.
+
+A custom authorizer has to be configured on the Kafka Broker to take advantage of Authorization Services REST endpoints available on Keycloak, which provide a list of granted permissions on resources for authenticated users.
+The list is fetched once, and enforced locally on the Kafka Broker for each user session in order to provide fast authorization decisions.
```
scholzj: > The list is fetched once

Is it once per user session? I.e. disconnecting it and connecting again would get a fresh list? Would some time based limit make more sense here? This could be a security issue for many users.

examples/README-authz.md
```diff
@@ -0,0 +1,403 @@
...
+Under `Users` click on `View all users` button to see two users defined - `alice` and `bob`. They are both members of some group. 
+In Keycloak the users can be stored in an LDAP identity provider.
+
+Under `Roles` we can see several realm roles which can be used to mark users or clients as having some permissions.
+Roles are a concept analogous to groups. One exception in Keycloak is that roles can not be stored in an LDAP identity provider.
+
+Under `Clients` we can see some additional clients configured - `kafka`, `kafka-cli`, `team-a-client`, `team-b-client`.
+The client with client id `kafka` is used by Kafka Brokers to perform the necessary OAuth2 communication for access token validation, 
+and to authenticate to other Kafka Broker instances using OAuth2 client authentication.
+This client also contains Authorization Services resource definitions, policies and authorization scopes used to perform authorization on Kafka Brokers.
```
scholzj: To many users, this does not mean anything. Can we get some clarity on what these are here?

examples/README-authz.md
```diff
@@ -0,0 +1,403 @@
...
+These are resource specifiers - patterns in a specific format, that are used to target policies to specific resources.
+The format is quite simple. For example:
+
+- `kafka-cluster:cluster-1,Topic:a_*`  ... targets only topics in kafka cluster 'cluster-1' with names starting with 'a_'
+
+If `kafka-cluster:XXX` segment is not present, the specifier targets any cluster.
+
+- `Group:x_*` ... targets all consumer groups on any cluster with names starting with 'x_'
+
+The possible resource types mirror the Kafka authorization model (Topic, Group, Cluster).
```
scholzj: The KAfka AuthZ model has a lot more than these 3. So can this work with all of them?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/JSONUtil.java
mstruk: I improved this a little bit as it turns out _asText()_ does not just automatically convert object node to json string.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
mstruk: There are now two additional logs - one for granted permissions and one for denied. One needs to enable debug logging on them to see logging output.

examples/README-authz.md
```diff
@@ -0,0 +1,353 @@
+## Token based authorization with Keycloak Authorization Services
+
+Once the Kafka Broker has obtained an access token by using Strimzi Kafka OAuth for authentication, it is possible to use centrally managed authorization rules to enforce access restrictions onto Kafka Clients.
+For this Strimzi Kafka OAuth supports the use of Keycloak Authorization Services.
+
+A custom authorizer has to be configured on the Kafka Broker to take advantage of Authorization Services REST endpoints available on Keycloak, which provide a list of granted permissions on resources for authenticated users.
+The list is fetched once, and enforced locally on Kafka Broker for each user session in order to provide fast authorization decisions.
+
+## Authorization example
```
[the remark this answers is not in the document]
mstruk: Did some segmantation into chapters.

examples/README-authz.md
mstruk: Looks good [here](https://github.com/mstruk/strimzi-kafka-oauth/blob/authz/examples/README-authz.md).

examples/README-authz.md
mstruk: Yes, it is once per connection (Kafka broker session) currently.

I guess adding periodic refresh is the next thing to add as @tombentley has also [expressed concern about this](https://github.com/strimzi/strimzi-kafka-oauth/pull/24#discussion_r367803753).

examples/README-authz.md
mstruk: These are somewhat abstract concepts, not entirely easy to explain. For starters we could add a link to [Keycloak Authorization Services docs](https://www.keycloak.org/docs/latest/authorization_services/index.html).
Maybe manage to put together a short explanation that would be enough to get started.

examples/README-authz.md
mstruk: Based on the [Kafka documentation](https://kafka.apache.org/documentation/#resources_in_kafka) the different resource types are: Topic, Group, Cluster, Transactionald, and DelegationToken. 

TransactionalId is supported, but I did not add support for DelegationToken since I imagine that's never used in conjunction with OAuth. If it is, it would require Authorizer to also work with non-oauth authentication, when there is no JWT token available.

scholzj: I'm not completely sure about the `DelegationToken`. But keep in mind this:
* There is only one authorizer per cluster
* There might be different interfaces with different auths and that is why we have the fallback option to the Simple authorizer.

So I think it is fine if we don't expect any `DelegationToken` rules in Keycloak. But there migth be usecases when we need to pass it to the fallback SImpleACLAuthorizer.

examples/README-authz.md
mstruk: @ppatierno What is the standard practice for tracking external issues 'here'? Creating a new Issue and linking to external one in its description, then just leaving it open and occasionally checking it?

examples/docker/kafka-oauth-strimzi/kafka/oauth.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash
+
+usage() {
+  echo "Usage: $0 [USERNAME] [PASSWORD] [ARGUMENTS] ..."
+  echo
+  echo "$0 is a tool for obtaining an access token or a refresh token for the user or the client."
```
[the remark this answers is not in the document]
tombentley: Yes, it would. IIRC we include the jar using the 3rd party jars mechanism, but we don't really have similar mechanism for including tools or scripts.

examples/docker/kafka-oauth-strimzi/kafka/oauth.sh
```diff
@@ -0,0 +1,121 @@
...
+  echo
+  echo " USERNAME    The username for user authentication"
+  echo " PASSWORD    The password for user authentication (prompted for if not specified)"
+  echo
+  echo " If USERNAME and PASSWORD are not specified, client credentials will be used for authentication."
```
tombentley: What are client credentials?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/JwtKafkaPrincipal.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    public boolean equals(Object o) {
+        // We don't care about jwt field for equals comparison
+        return super.equals(o);
+    }
+
+    @Override
+    public int hashCode() {
+        return super.hashCode();
+    }
```
tombentley: Why are we overriding these if we're just delegating to super?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
```diff
@@ -0,0 +1,436 @@
...
+
+    @Override
+    public scala.collection.immutable.Map<Resource, Set<Acl>> getAcls(KafkaPrincipal principal) {
+        return super.getAcls(principal);
+    }
+
+    @Override
+    public scala.collection.immutable.Map<Resource, Set<Acl>> getAcls() {
+        return super.getAcls();
+    }
```
tombentley: By delegating these to super you're effectively letting people use the Kafka Admin `createAcls()`, `alterAcls()` and `deleteAcls()` even when `delegateToKafkaACL==false`. That's a bit misleading. Maybe we shouldn't go as far as to throw some exception in that case, but I certainly think some WARN logging is in order.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/UserSpec.java
```diff
@@ -0,0 +1,34 @@
...
+        return type;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+
+    public static UserSpec of(String principal) {
+        String[] spec = principal.split(":");
```
tombentley: Maybe use limit, just in case the name includes `:` (though I'm sure you'll tell me that's impossible)?

examples/docker/kafka-oauth-strimzi/kafka/oauth.sh
mstruk: `--client-id` and `--secret`.

tombentley: I think you should say the explicitly.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/JwtKafkaPrincipal.java
mstruk: Maybe just my lazy way of dealing with SpotBugs. I would actually prefer to add SpotBugs exception. Not yet quite versed in how to decide each time.
I'll gladly add SpotBugs exception.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/UserSpec.java
mstruk: That's a good catch. We could play super safe and not assume anything about usernames ...

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/JwtKafkaPrincipal.java
tombentley: Maybe I'm being stupid: Does the superclass do an `instanceof` or does it check class equality?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
mstruk: That's a fair point. Honestly, I don't quite know what to do with these. 
We could go with what you suggest, and throw exception if delegating to simple ACL is not enabled.

tombentley: Yeah, either exception or logging. @ppatierno @scholzj any preference?

ppatierno: No strong opinion on my side, I am fine with both.

tombentley: The only use case I can see for allowing them to create, alter and delete ACLs if they're not using them would be if they were trying to, e.g. delete ACLs after moving to using authz services (or setting them up when moving the other way). That seems like quite a stretch.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/JwtKafkaPrincipal.java
mstruk: In fact it's not a good practice to do any assumptions about super equals() that is not part of our code base. I'll do a proper equals impl.

mstruk:
I have some more fixes coming, for issues identified while testing with mutual TLS listeners and Kafka 2.4 ...

examples/README-authz.md
ppatierno: @mstruk yes, it's the only way I know :-/

examples/README-authz.md
mstruk: I created several chapter. This one I called 'Authorization in Action Using CLI Clients'

examples/README-authz.md
mstruk: Yes, this is for the benefit of the reader, reading MarkDown on GitHub, so that the lines don't get truncated by viewport.

examples/README-authz.md
mstruk: I added all in the code, for completeness.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
mstruk: I made it so that exception is thrown if delegation to simple Kafka ACL isn't enabled.

oauth-keycloak-authorizer/etc/authorization-scopes.json
```diff
@@ -0,0 +1,37 @@
+{
```
scholzj: Could you please explain why do we need this file here? How is it used and how is it packaged?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/AuthzConfig.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package io.strimzi.kafka.oauth.server.authorizer;
+
+import io.strimzi.kafka.oauth.common.Config;
+
+import java.util.Properties;
+
+public class AuthzConfig extends Config {
+
+    public static final String STRIMZI_AUTHZ_CLIENT_ID = "strimzi.authz.client.id";
```
scholzj: I wonder if we should use `authorization` intead of `authz`. Lot of people does not know / use this as shortcut and might be confused.

oauth-keycloak-authorizer/etc/authorization-scopes.json
mstruk: Currently all the documentation is in [examples/docker/README-authz.md](https://github.com/mstruk/strimzi-kafka-oauth/blob/authz/examples/README-authz.md) which is a walkthrough example. This file is mentioned there. I can add a paragraph that explains how to import it into empty Authorization Services configuration.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/AuthzConfig.java
mstruk: Fixed.

oauth-keycloak-authorizer/etc/authorization-scopes.json
mstruk: Fixed.

mstruk:
It looks like the tests could use a longer timeout to give enough time to pull down images from docker hub.

PR #25, opened 2020-01-15
Provide EC signature verification support in JWTSignatureValidator

mstruk:
This PR attempts to address: https://github.com/strimzi/strimzi-kafka-operator/issues/2327

### DIFF
--- a/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/ECDSASignatureVerifierContext.java
+++ b/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/ECDSASignatureVerifierContext.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2017-2020, Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.oauth.validator;
+
+import org.bouncycastle.asn1.ASN1Integer;
+import org.bouncycastle.asn1.DERSequenceGenerator;
+import org.keycloak.common.VerificationException;
+import org.keycloak.crypto.AsymmetricSignatureVerifierContext;
+import org.keycloak.crypto.KeyWrapper;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.math.BigInteger;
+
+/**
+ * This class provides EC signature verification support.
+ *
+ * Adapted from: https://github.com/keycloak/keycloak/blob/8.0.1/services/src/main/java/org/keycloak/crypto/ClientECDSASignatureVerifierContext.java
+ */
+public class ECDSASignatureVerifierContext extends AsymmetricSignatureVerifierContext {
+
+    public ECDSASignatureVerifierContext(KeyWrapper key) {
+        super(key);
+    }
+
+    @Override
+    public boolean verify(byte[] data, byte[] signature) throws VerificationException {
+       /*
+       Fallback for backwards compatibility of ECDSA signed tokens which were issued in previous versions.
+       TODO remove by https://issues.jboss.org/browse/KEYCLOAK-11911
+        */
+        int expectedSize = ECDSA.valueOf(getAlgorithm()).getSignatureLength();
+        byte[] derSignature = expectedSize != signature.length && signature[0] == 0x30 ? signature : concatenatedRSToASN1DER(signature, expectedSize);
+        return super.verify(data, derSignature);
+    }
+
+    enum ECDSA {
+        ES256(64),
+        ES384(96),
+        ES512(132);
+
+        private final int signatureLength;
+
+        ECDSA(int signatureLength) {
+            this.signatureLength = signatureLength;
+        }
+
+        public int getSignatureLength() {
+            return this.signatureLength;
+        }
+    }
+
+    static byte[] concatenatedRSToASN1DER(final byte[] signature, int signLength) {
+        int len = signLength / 2;
+        int arraySize = len + 1;
+
+        byte[] r = new byte[arraySize];
+        byte[] s = new byte[arraySize];
+        System.arraycopy(signature, 0, r, 1, len);
+        System.arraycopy(signature, len, s, 1, len);
+        BigInteger rBigInteger = new BigInteger(r);
+        BigInteger sBigInteger = new BigInteger(s);
+
+        ByteArrayOutputStream bos = new ByteArrayOutputStream();
+        try {
+            DERSequenceGenerator seqGen = new DERSequenceGenerator(bos);
+
+            seqGen.addObject(new ASN1Integer(rBigInteger.toByteArray()));
+            seqGen.addObject(new ASN1Integer(sBigInteger.toByteArray()));
+            seqGen.close();
+            bos.close();
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to generate ASN.1 DER signature", e);
+        }
+        return bos.toByteArray();
+    }
+}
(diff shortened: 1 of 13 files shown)


oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
```diff
@@ -36,6 +40,10 @@
 
 public class JWTSignatureValidator implements TokenValidator {
 
+    static {
+        Security.insertProviderAt(new BouncyCastleProvider(), 1);
```
tombentley: This will add BC as a security provider for a whole bunch of algorithms, not just ECDSA, right? And we don't know whether the user of the library trusts BC? So maybe we should only register when we encounter a token using `EC` and only if there's not existing EC provision already registered? That would allow a user who really cares to register some other provider for EC and we'd just pick it up.

Also, I don't think we can assume that there aren't other providers already registered in some specific order. Adding at index 1 might then mean BC is chosen instead of their preferred provider. So maybe if/when we do add BC, we should always add at the end (which is still a bit racy, but less likely to be a problem in practice)?

testsuite/client-secret-jwt-keycloak-test/src/test/java/io/strimzi/testsuite/oauth/KeycloakClientCredentialsWithJwtValidationTest.java
```diff
@@ -30,11 +30,11 @@
 public class KeycloakClientCredentialsWithJwtValidationTest {
 
     private static final String HOST = "keycloak";
-    private static final String REALM = "demo";
+    private static final String REALM = "demo-ec";
```
tombentley: Don't we need to be able to test both EC and non-EC?

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
mstruk: Right. That is a dilemma. The thing is that for EC, the Keycloak helper library that we use requests ECDSA which is provided by Bouncycastle, but is otherwise not available in JDK. To make make matters worse, there's usage of both "EC" and "ECDSA" in the code. That means if Bouncycastle is not installed in first place what will happen is that when "EC" is requested it would use Sun's provider, and when "ECDSA" is requested the Bouncycastle provider would be used - unless the provider is bumped to first position among security providers. Multiple impls might work together fine or they might not.

We could add a configuration: `oauth.crypto.enable.bouncycastle` to require explicitly enabling the provider before we install it. And another - `oauth.crypto.bouncycastle.provider.position` which, if set to 0 or negative number would install the provider at the end, otherwise at the position specified - or at the end if position is beyond number of providers.

I'm not sure about lazy enabling it upon detection of EC. That may hide any potential issues to long after Kafka broker has started.

testsuite/client-secret-jwt-keycloak-test/src/test/java/io/strimzi/testsuite/oauth/KeycloakClientCredentialsWithJwtValidationTest.java
mstruk: The default used is RSA which is used in all the other tests. But this specific test I adjusted to use EC. It may be a bit obfuscated, but both the default and EC get tested over the course of a testsuite run. These tests take some time and some hardware resources to run, so I try to avoid doing permutations.

tombentley: OK, so long as we haven't lost all coverage of RSA that's fine.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
tombentley: I'm guessing that the helper library has no mechanism for selecting algorithm impls with a given provider?

It not really clear to me whether the "EC" signature algo is actually the same thing as "ECDSA" (in which case Keycloak's use of EC starts to look like a bug). But I guess we need to provide something which works with the version of the library which actually exists. So I suppose System properties like `oauth.crypto.bouncycastle.provider.enable` would at least provide a mechanism, even if it's less than ergonomic.

mstruk: Right, the Keycloak code is what it is, and it does not explicitly request BouncyCastle provider. With the [patched Keycloak library](https://github.com/keycloak/keycloak/pull/6643) that requests "EC" instead of "ECDSA" things work fine without BouncyCastle. But that would require a patch that probably won't make it into the master, and will only be released later while we would like a fix ASAP.

mstruk:
I made installing BouncyCastle provider optional. It is required for EC signature verification to work, though.

It can be installed as the last security provider by passing to JAAS config of `org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule` the parameter:

    oauth.crypto.provider.bouncycastle="true"

Position in the providers list can be controlled with another parameter:

    oauth.crypto.provider.bouncycastle.position="0"

Alternatively, these can be set as a system properties or as ENV vars called `OAUTH_CRYPTO_PROVIDER_BOUNCYCASTLE` and `OAUTH_CRYPTO_PROVIDER_BOUNCYCASTLE_POSITION`.

scholzj:
@ppatierno Do you plan to review this or can it be merged?

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/ECDSASignatureVerifierContext.java
```diff
@@ -0,0 +1,79 @@
...
+import org.keycloak.common.VerificationException;
+import org.keycloak.crypto.AsymmetricSignatureVerifierContext;
+import org.keycloak.crypto.KeyWrapper;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.math.BigInteger;
+
+/**
+ * This class provides EC signature verification support.
```
ppatierno: ECDSA not EC

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
```diff
@@ -97,6 +107,20 @@ public JWTSignatureValidator(String keysEndpointUri,
...
+            int installedPosition = Security.insertProviderAt(new BouncyCastleProvider(), bouncyCastleProviderPosition);
+            log.info("BouncyCastle security provider installed at position: " + installedPosition);
+
+            if (log.isDebugEnabled()) {
+                StringBuilder sb = new StringBuilder("Installed security providers:\n");
+                for (Provider p: Security.getProviders()) {
+                    sb.append("  - " + p.toString() + "  [" + p.getClass().getName() + "]\n");
+                    sb.append("   " + p.getInfo() + "\n");
+                }
+                log.debug(sb.toString());
```
ppatierno: this should log all the installed security providers but it happens only if the outer `if (enableBouncyCastleProvider && !bouncyInstalled.getAndSet(true))` is true. Is it really what we want? Shouldn't we log the installed security providers anyway?

ppatierno:
@scholzj I approved but left a couple of comments. Maybe changes are not needed but I will let @mstruk to decide.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
mstruk: This is the way I intended. Usually you're not really interested in security providers, you're working with what's packaged and enabled in JDK. If you really want to see the providers for some reason, you can just enable bouncy castle provider in the config, and it will be added at the end, not overwriting any existing providers, but you'll also see them all listed.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/ECDSASignatureVerifierContext.java
mstruk: Fixed.

PR #30, opened 2020-02-11
Make OAuth scope optional

klalafaryan:
Some OAuth Authorization servers such as AWS Cognito don't provide the `scope` in the response.  For example:
```
{
    "access_token": "{JWT_TOKEN}",
    "expires_in": 3600,
    "token_type": "Bearer"
}
```

This PR  makes `scope` property optional which gives us possibility to integrate with AWS Cognito. (Authorization servers which are not providing `scope` in the response)

### DIFF
--- a/oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
+++ b/oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
@@ -102,7 +102,8 @@ public class OAuthAuthenticator {
             throw new IllegalStateException("Invalid response from authorization server: no expires_in");
         }
 
-        // Some OAuth2 authorization servers don't provide scope in this level
+        // Some OAuth2 authorization servers don't provide scope in this level,
+        // therefore we don't need to make it mandatory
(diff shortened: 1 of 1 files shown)


mstruk:
The fix is in fact on the Kafka client (the client side of OAuth) - it doesn't really concern the Kafka Broker unless inter-broker communication goes over OAuth protected listener (which in Strimzi Kafka Operator it doesn't).

I think adding the comment in the code is unnecessary, but overall the PR looks good. 

@klalafaryan Thank you for your contribution.

mstruk:
@ppatierno I guess we won't test it immediately - the addressed situation never occurs in our current test setups (using Keycloak and Hydra). We could create simple authorization server endpoints that would do just enough so we can test this, eventually.

scholzj:
@mstruk So we do not use the scope for anything in the broker part?

mstruk:
@scholzj Correct. Authentication part does not use it - it's more for authorization use, but there SimpleACL isn't aware of it, and KeycloakRBACAuthorizer doesn't make use of it either - in favor of fine grained security provided by Authorization Services.

scholzj:
Thanks for the PR.

PR #36, opened 2020-03-11
Align configuration with Kafka Operator

mstruk:
This PR introduces changes with regards to configuration:
* `oauth.tokens.not.jwt` is now called `oauth.access.token.is.jwt` and has a reverse meaning
* `oauth.validation.skip.type.check` is now called `oauth.check.access.token.type` and has a reverse meaning

Having a reverse meaning for both these configuration parameters means that the default value is `true` whereas before it was `false`, and when setting them explicitly you usually want to set them to `false` rather than `true` as it was the case before.

The old configuration parameter names still work with the old semantics, but have been marked @Deprecated and will be removed in future versions. A Warning is logged if the old parameter names are used.

### DIFF
--- a/examples/consumer/src/main/java/io/strimzi/examples/consumer/ExampleConsumer.java
+++ b/examples/consumer/src/main/java/io/strimzi/examples/consumer/ExampleConsumer.java
@@ -55,7 +55,7 @@ public class ExampleConsumer {
         }
 
         // Use 'preferred_username' rather than 'sub' for principal name
-        if (!external.getValueAsBoolean(Config.OAUTH_TOKENS_NOT_JWT, false)) {
+        if (external.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true)) {
             defaults.setProperty(Config.OAUTH_USERNAME_CLAIM, "preferred_username");
         }
 
--- a/examples/consumer/src/main/java/io/strimzi/examples/consumer/ExampleConsumer.java
+++ b/examples/consumer/src/main/java/io/strimzi/examples/consumer/ExampleConsumer.java
@@ -55,7 +55,7 @@ public class ExampleConsumer {
         }
 
         // Use 'preferred_username' rather than 'sub' for principal name
-        if (external.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true)) {
+        if (isAccessTokenJwt(external)) {
             defaults.setProperty(Config.OAUTH_USERNAME_CLAIM, "preferred_username");
         }
 
@@ -73,6 +73,16 @@ public class ExampleConsumer {
                 System.out.println("Consumed message: " + record.value());
             }
         }
+    }
+
+    @SuppressWarnings("deprecation")
+    private static boolean isAccessTokenJwt(Config config) {
+        String legacy = config.getValue(Config.OAUTH_TOKENS_NOT_JWT);
+        if (legacy != null) {
+            System.out.println("[WARN] OAUTH_TOKENS_NOT_JWT is deprecated. Use OAUTH_ACCESS_TOKEN_IS_JWT (with reverse meaning) instead.");
+        }
+        return legacy != null ? !Config.isTrue(legacy) :
+                config.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true);
     }
 
     /**
--- a/examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
+++ b/examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
@@ -53,7 +53,7 @@ public class ExampleProducer {
         }
 
         // Use 'preferred_username' rather than 'sub' for principal name
-        if (external.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true)) {
+        if (isAccessTokenJwt(external)) {
             defaults.setProperty(Config.OAUTH_USERNAME_CLAIM, "preferred_username");
         }
 
@@ -84,6 +84,16 @@ public class ExampleProducer {
                 throw new RuntimeException("Interrupted while sleeping!");
             }
         }
+    }
+
+    @SuppressWarnings("deprecation")
+    private static boolean isAccessTokenJwt(Config config) {
+        String legacy = config.getValue(Config.OAUTH_TOKENS_NOT_JWT);
+        if (legacy != null) {
+            System.out.println("[WARN] OAUTH_TOKENS_NOT_JWT is deprecated. Use OAUTH_ACCESS_TOKEN_IS_JWT (with reverse meaning) instead.");
+        }
+        return legacy != null ? !Config.isTrue(legacy) :
+                config.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true);
     }
 
     /**
--- a/examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
+++ b/examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
@@ -53,7 +53,7 @@ public class ExampleProducer {
         }
 
         // Use 'preferred_username' rather than 'sub' for principal name
-        if (!external.getValueAsBoolean(Config.OAUTH_TOKENS_NOT_JWT, false)) {
+        if (external.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true)) {
             defaults.setProperty(Config.OAUTH_USERNAME_CLAIM, "preferred_username");
         }
 
--- a/oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
+++ b/oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
@@ -99,7 +99,7 @@ public class JaasClientOauthLoginCallbackHandler implements AuthenticateCallback
             usernameClaim = null;
         }
 
-        isJwt = config.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true);
+        isJwt = isAccessTokenJwt(config);
         if (!isJwt && usernameClaim != null) {
             throw new RuntimeException("Custom username claim (OAUTH_USERNAME_CLAIM) not available, when tokens are configured as opaque (OAUTH_ACCESS_TOKEN_IS_JWT=false)");
         }
@@ -119,6 +119,16 @@ public class JaasClientOauthLoginCallbackHandler implements AuthenticateCallback
                     + "\n    maxTokenExpirySeconds: " + maxTokenExpirySeconds
                     + "\n    usernameClaim: " + usernameClaim);
         }
+    }
+
+    @SuppressWarnings("deprecation")
+    private static boolean isAccessTokenJwt(Config config) {
+        String legacy = config.getValue(Config.OAUTH_TOKENS_NOT_JWT);
+        if (legacy != null) {
+            log.warn("OAUTH_TOKENS_NOT_JWT is deprecated. Use OAUTH_ACCESS_TOKEN_IS_JWT (with reverse meaning) instead.");
+        }
+        return legacy != null ? !Config.isTrue(legacy) :
+                config.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true);
     }
 
     @Override
--- a/oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
+++ b/oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
@@ -45,7 +45,7 @@ public class JaasClientOauthLoginCallbackHandler implements AuthenticateCallback
     private String clientSecret;
     private URI tokenEndpoint;
     private String usernameClaim;
-    private boolean isJWT;
+    private boolean isJwt;
     private int maxTokenExpirySeconds;
 
     private SSLSocketFactory socketFactory;
@@ -99,9 +99,9 @@ public class JaasClientOauthLoginCallbackHandler implements AuthenticateCallback
             usernameClaim = null;
         }
 
-        isJWT = !config.getValueAsBoolean(Config.OAUTH_TOKENS_NOT_JWT, false);
-        if (!isJWT && usernameClaim != null) {
-            throw new RuntimeException("Custom username claim (OAUTH_USERNAME_CLAIM) not available, when tokens are configured as opaque (OAUTH_TOKENS_NOT_JWT)");
+        isJwt = config.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true);
+        if (!isJwt && usernameClaim != null) {
+            throw new RuntimeException("Custom username claim (OAUTH_USERNAME_CLAIM) not available, when tokens are configured as opaque (OAUTH_ACCESS_TOKEN_IS_JWT=false)");
         }
 
         maxTokenExpirySeconds = config.getValueAsInt(ClientConfig.OAUTH_MAX_TOKEN_EXPIRY_SECONDS, -1);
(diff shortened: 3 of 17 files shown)


tombentley:
I know this is still 0.x stuff, but I don't think it's necessary to make breaking changes. I think it would be better to deprecate `oauth.tokens.not.jwt` (and log a warning if it's used), and error if `oauth.tokens.not.jwt` and `oauth.tokens.is.jwt` are both set. Ditto the other config. Why make life harder for your users? Let them fix their configs in their own time, and don't assume they're going to read the release notes. You can remove the deprecated configs when you hit 1.0, which is when users are more likely to actually read the release notes.

mstruk:
@tombentley I made the change as you suggested. Old config options are still in place, working as they used to, but marked as deprecated, and a warning is logged if they are used.

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
```diff
@@ -115,12 +115,22 @@ public void configure(Map<String, ?> configs, String saslMechanism, List<AppConf
...
 
+    @SuppressWarnings("deprecation")
+    private static boolean isAccessTokenJwt(Config config) {
+        String legacy = config.getValue(Config.OAUTH_TOKENS_NOT_JWT);
+        if (legacy != null) {
+            log.warn("OAUTH_TOKENS_NOT_JWT is deprecated. Use OAUTH_ACCESS_TOKEN_IS_JWT (with reverse meaning) instead.");
+        }
+        return legacy != null ? !Config.isTrue(legacy) :
+                config.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true);
+    }
```
tombentley: Same two comments

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
```diff
@@ -113,6 +113,26 @@ public void configure(Map<String, ?> configs, String saslMechanism, List<AppConf
...
+
+    @SuppressWarnings("deprecation")
+    private static boolean isCheckAccessTokenType(Config config) {
+        String legacy = config.getValue(ServerConfig.OAUTH_VALIDATION_SKIP_TYPE_CHECK);
+        if (legacy != null) {
+            log.warn("OAUTH_VALIDATION_SKIP_TYPE_CHECK is deprecated. Use OAUTH_CHECK_ACCESS_TOKEN_TYPE (with reverse meaning) instead.");
+        }
+        return legacy != null ? !Config.isTrue(legacy) :
+                config.getValueAsBoolean(ServerConfig.OAUTH_CHECK_ACCESS_TOKEN_TYPE, true);
+    }
```
tombentley: So at this point I have to wonder why you couldn't write these methods once and reuse them? Surely you have some `Utils` class?

examples/consumer/src/main/java/io/strimzi/examples/consumer/ExampleConsumer.java
```diff
@@ -75,6 +75,16 @@ public static void main(String[] args) {
         }
     }
 
+    @SuppressWarnings("deprecation")
+    private static boolean isAccessTokenJwt(Config config) {
+        String legacy = config.getValue(Config.OAUTH_TOKENS_NOT_JWT);
+        if (legacy != null) {
+            System.out.println("[WARN] OAUTH_TOKENS_NOT_JWT is deprecated. Use OAUTH_ACCESS_TOKEN_IS_JWT (with reverse meaning) instead.");
```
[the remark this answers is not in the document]
mstruk: These is an example consumer meant to be used from CLI. Not using logging is on purpose.

examples/consumer/src/main/java/io/strimzi/examples/consumer/ExampleConsumer.java
```diff
@@ -75,6 +75,16 @@ public static void main(String[] args) {
...
+        }
+        return legacy != null ? !Config.isTrue(legacy) :
+                config.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true);
+    }
```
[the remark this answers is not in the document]
mstruk: Actually, using both config options at the same time should rather be treated as a configuration error, and exception thrown IMHO. Having them both set can not just happen by deploying a new version of code onto old configuration - one also has to change config. So I don't think this falls into 'backwards compatibility' category.

examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
```diff
@@ -86,6 +86,16 @@ public static void main(String[] args) {
...
     }
 
+    @SuppressWarnings("deprecation")
+    private static boolean isAccessTokenJwt(Config config) {
+        String legacy = config.getValue(Config.OAUTH_TOKENS_NOT_JWT);
+        if (legacy != null) {
+            System.out.println("[WARN] OAUTH_TOKENS_NOT_JWT is deprecated. Use OAUTH_ACCESS_TOKEN_IS_JWT (with reverse meaning) instead.");
+        }
+        return legacy != null ? !Config.isTrue(legacy) :
+                config.getValueAsBoolean(Config.OAUTH_ACCESS_TOKEN_IS_JWT, true);
```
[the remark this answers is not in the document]
mstruk: Same two answers :)

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
mstruk: isAccessTokenJwt() could indeed be put into some common class. Good catch.

examples/consumer/src/main/java/io/strimzi/examples/consumer/ExampleConsumer.java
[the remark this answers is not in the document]
tombentley: Fine with me.

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
```diff
@@ -99,9 +100,9 @@ public void configure(Map<String, ?> configs, String saslMechanism, List<AppConf
             usernameClaim = null;
         }
 
-        isJWT = !config.getValueAsBoolean(Config.OAUTH_TOKENS_NOT_JWT, false);
-        if (!isJWT && usernameClaim != null) {
-            throw new RuntimeException("Custom username claim (OAUTH_USERNAME_CLAIM) not available, when tokens are configured as opaque (OAUTH_TOKENS_NOT_JWT)");
+        isJwt = isAccessTokenJwt(config, log, null);
+        if (!isJwt && usernameClaim != null) {
+            throw new RuntimeException("Custom username claim (OAUTH_USERNAME_CLAIM) not available, when tokens are configured as opaque (OAUTH_ACCESS_TOKEN_IS_JWT=false)");
```
tombentley: Should we have a more specific exception type for this?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/DeprecationUtil.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class DeprecationUtil {
+
+    @SuppressWarnings("deprecation")
+    public static boolean isAccessTokenJwt(Config config, Logger log, String errorPrefix) {
+        String legacy = config.getValue(Config.OAUTH_TOKENS_NOT_JWT);
+        if (legacy != null) {
+            log.warn("OAUTH_TOKENS_NOT_JWT is deprecated. Use OAUTH_ACCESS_TOKEN_IS_JWT (with reverse meaning) instead.");
+            if (config.getValue(Config.OAUTH_ACCESS_TOKEN_IS_JWT) != null) {
+                throw new RuntimeException((errorPrefix != null ? errorPrefix : "") + "Can't use both OAUTH_ACCESS_TOKEN_IS_JWT and OAUTH_TOKENS_NOT_JWT");
```
tombentley: More specific exception type, maybe? Wdyt?

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
mstruk: We don't handle this exception anywhere. Any exception thrown here is caught by Kafka Broker and results in broker shutting down. 

I don't think there is some Kafka runtime exception that _should_ be used here? If that were the case we should replace with it all occurrences of RuntimeException inside configure() method.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/DeprecationUtil.java
mstruk: Same answer as above.

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
tombentley: I'm just thinking that if your Kafka broker shuts down in these circumstances it's slightly nicer to see `OauthConfigurationException` in the logs than `RuntimeException`.

mstruk: That's very subjective. I personally don't see that as added value. But since it would require changing all RuntimeException throw statements in configure(), if we really want that, I prefer we make a new PR just for that and keep this one focused on configuration alignment. Maybe create a JIRA for the other one if you think we really should introduce a new exception.

examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
```diff
@@ -86,6 +86,16 @@ public static void main(String[] args) {
         }
     }
 
+    @SuppressWarnings("deprecation")
+    private static boolean isAccessTokenJwt(Config config) {
+        String legacy = config.getValue(Config.OAUTH_TOKENS_NOT_JWT);
+        if (legacy != null) {
+            System.out.println("[WARN] OAUTH_TOKENS_NOT_JWT is deprecated. Use OAUTH_ACCESS_TOKEN_IS_JWT (with reverse meaning) instead.");
+        }
+        return legacy != null ? !Config.isTrue(legacy) :
```
scholzj: Same as above. Is it really correct that the legacy option is prefered over the new one? Shouldn't it be the other way around? Or does it have some other reason?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/DeprecationUtil.java
```diff
@@ -0,0 +1,24 @@
...
+            }
+        }
+
+        return legacy != null ? !Config.isTrue(legacy) :
```
scholzj: Same as above ... Is it really correct that the legacy option is prefered over the new one? Shouldn't it be the other way around? Or does it have some other reason?

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
```diff
@@ -113,6 +113,19 @@ public void configure(Map<String, ?> configs, String saslMechanism, List<AppConf
...
+    @SuppressWarnings("deprecation")
+    private static boolean isCheckAccessTokenType(Config config) {
+        String legacy = config.getValue(ServerConfig.OAUTH_VALIDATION_SKIP_TYPE_CHECK);
+        if (legacy != null) {
+            log.warn("OAUTH_VALIDATION_SKIP_TYPE_CHECK is deprecated. Use OAUTH_CHECK_ACCESS_TOKEN_TYPE (with reverse meaning) instead.");
+            if (config.getValue(ServerConfig.OAUTH_CHECK_ACCESS_TOKEN_TYPE) != null) {
+                throw new RuntimeException("OAuth validator configuration error: can't use both OAUTH_CHECK_ACCESS_TOKEN_TYPE and OAUTH_VALIDATION_SKIP_TYPE_CHECK");
+            }
+        }
+        return legacy != null ? !Config.isTrue(legacy) :
```
scholzj: Same as above ... Is it really correct that the legacy option is prefered over the new one? Shouldn't it be the other way around? Or does it have some other reason?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
```diff
@@ -128,7 +131,7 @@ public boolean getValueAsBoolean(String key, boolean fallback) {
         }
     }
 
-    private boolean isTrue(String result) {
+    public static boolean isTrue(String result) {
```
ppatierno: of course this was already here but why we are supporting all these possible values and not just "true/false"?

examples/consumer/src/main/java/io/strimzi/examples/consumer/ExampleConsumer.java
```diff
@@ -75,6 +75,16 @@ public static void main(String[] args) {
         }
     }
 
+    @SuppressWarnings("deprecation")
+    private static boolean isAccessTokenJwt(Config config) {
+        String legacy = config.getValue(Config.OAUTH_TOKENS_NOT_JWT);
+        if (legacy != null) {
+            System.out.println("[WARN] OAUTH_TOKENS_NOT_JWT is deprecated. Use OAUTH_ACCESS_TOKEN_IS_JWT (with reverse meaning) instead.");
+        }
+        return legacy != null ? !Config.isTrue(legacy) :
```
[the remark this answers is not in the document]
mstruk: This is just the ExampleConsumer CLI client - not something anyone will use for more than learning, how to use it on the client. Validation performed by it is more for the user experience than for validation. 
Actual validation happens later in the workflow. If both old and new options are present, then the exception is thrown, because it's treated as a misconfiguration.
It happens in [OAuth client library](https://github.com/mstruk/strimzi-kafka-oauth/blob/backport-config/oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java#L103) that actually handles this config under the hood and [throws an exception](https://github.com/mstruk/strimzi-kafka-oauth/blob/backport-config/oauth-common/src/main/java/io/strimzi/kafka/oauth/common/DeprecationUtil.java#L17). 

So, if both options are present then the exception will be thrown, just a little later in the code flow.

examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
mstruk: Same answer.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/DeprecationUtil.java
mstruk: This code won't be reached if both are present. See [previous lines](https://github.com/mstruk/strimzi-kafka-oauth/blob/backport-config/oauth-common/src/main/java/io/strimzi/kafka/oauth/common/DeprecationUtil.java#L17)

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
mstruk: Same answer - this code won't be reached if both are present.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
mstruk: The only reason would be usability / user experience. Some people may in some situation prefer 'y' over 'true', or '0' over 'false'. It will be a string in input in any case, and needs to be converted to boolean somehow. This code here makes it both versatile, and specific - as we can always point to the algorithm with a [url](https://github.com/strimzi/strimzi-kafka-oauth/blob/master/oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java#L131-L140).

@ppatierno Do you see some negatives of this choice?

scholzj:
> It would be nice to mention these deprecations yes. Funny we don't have CHANGELOG.md yet or RELEASE_NOTES.md. We had no place to put the mention of the new features until now :)

Yeah, so I think the point of my comment was _should add it here in this PR?_ Or should we merge it without it. :-)

> We also have to be careful in strimzi-kafka-operator to not pull in the next release without an accompanying strimzi-kafka-operator PR that rewires to these new properties, so that users won't see deprecation messages in the log. I have the code ready, just didn't open the PR yet.

I think we can do the release soon if you want - up to you.

mstruk:
Ok, I can add CHANGELOG.md, and put in it the changes for the previous releases as well.

I'm not in a hurry with the release I think - 0.17.0 train has left for this one so I don't see any hurry.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
ppatierno: in general, I don't like to give too much freedom to users :-)
Anyway I can live with it.

mstruk:
Instead of CHANGELOG.md I added RELEASE_NOTES.md as part of a different PR (#34, #44)

PR #39, opened 2020-03-17
Fix enabled remote debugging mode in example compose-authz.yml

mstruk:

### DIFF
--- a/examples/docker/kafka-oauth-strimzi/compose-authz.yml
+++ b/examples/docker/kafka-oauth-strimzi/compose-authz.yml
@@ -11,14 +11,14 @@ services:
       - 9092:9092
 
       # javaagent debug port
-      - 5006:5006
+      #- 5006:5006
 
     environment:
 
       # Java Debug
-      KAFKA_DEBUG: y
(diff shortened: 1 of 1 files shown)


examples/docker/kafka-oauth-strimzi/compose-authz.yml
```diff
@@ -11,14 +11,14 @@ services:
...
 
     environment:
 
       # Java Debug
-      KAFKA_DEBUG: y
-      DEBUG_SUSPEND_FLAG: y
-      JAVA_DEBUG_PORT: 5006
+      #KAFKA_DEBUG: y
+      #DEBUG_SUSPEND_FLAG: y
+      #JAVA_DEBUG_PORT: 5006
```
ppatierno: maybe we need a comment like "uncomment following lines for enabling debugging ... "

mstruk: I think it should be obvious to everyone that removing hashes would activate these settings.
And I prefer to not add comments for obvious things.

PR #55, opened 2020-05-07
Fix scope support when refreshing the token + enhance examples and tests

mstruk:
The newly added `scope` support allows setting the `scope` parameter in a token request, which wasn't possible before. This PR addresses a tiny omission in initial `scope` support where authenticating with a refresh token (by setting oauth.refresh.token) didn't take `scope` configuration into account.

### DIFF
--- a/examples/kubernetes/kafka-oauth-single-spring.yaml
+++ b/examples/kubernetes/kafka-oauth-single-spring.yaml
@@ -1,0 +1,49 @@
+apiVersion: kafka.strimzi.io/v1beta1
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    version: 2.4.1
+    replicas: 1
+    listeners:
+      plain:
+        authentication:
+          type: oauth
+          introspectionEndpointUri: http://spring:8080/oauth/check_token
+          clientId: kafka
+          clientSecret:
+            secretName: my-cluster-kafka-client-secret
+            key: secret
+          checkIssuer: false
+          accessTokenIsJwt: false
+          userNameClaim: user_name
+          fallbackUserNameClaim: client_id
+          fallbackUserNamePrefix: client-account-
+      tls: {}
+    logging:
+      type: inline
+      loggers:
+        log4j.logger.io.strimzi: "DEBUG"
+    config:
+      offsets.topic.replication.factor: 1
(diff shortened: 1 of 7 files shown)


examples/kubernetes/kafka-oauth-single-spring.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: kafka.strimzi.io/v1beta1
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    version: 2.4.1
```
scholzj, line 7: Should this have 2.5.0 as well?

examples/kubernetes/kafka-oauth-single-spring.yaml
```diff
@@ -0,0 +1,49 @@
...
+      tls: {}
+    logging:
+      type: inline
+      loggers:
+        log4j.logger.io.strimzi: "DEBUG"
+    config:
+      offsets.topic.replication.factor: 1
+      transaction.state.log.replication.factor: 1
+      transaction.state.log.min.isr: 1
+      log.message.format.version: "2.4"
```
scholzj, line 32: Should this have 2.5 as well?

scholzj:
This should be cherry-picked for 0.5.0 for RC2.

examples/kubernetes/kafka-oauth-single-spring.yaml
mstruk: I like to use different versions to increase the probability of catching compatibility problems that we may have missed. This should work with versions 2.4.0, 2.4.1, and 2.5.0, but not 2.3.0 because the latest image for that would not contain the latest operator any more.
The whole decision which of the supported versions to use in this place is rather non-deterministic :)
One benefit of using 2.5.0 might be to prod users to use the latest. OTOH, using older version makes it clear that you don't have to use the latest if for some reason you don't want to or you can't.

scholzj: Ok - If that was intentional to keep it as 2.4.1 I'm fine with it. It was not really clear from the PR. Thanks.

scholzj:
@strimzi/maintainers Can someone please review this PR with the bug fix found by Marko? Once approved, I can do the RC2 with this fix.

PR #63, opened 2020-06-08
Add maven profiles to testsuite for running with different versions of Kafka images

mstruk:


.travis/build.sh
```diff
@@ -29,7 +29,9 @@ if [ ${JAVA_MAJOR_VERSION} -eq 1 ] ; then
   docker pull oryd/hydra:v1.0.0
   mvn test-compile spotbugs:check -e -V -B -f testsuite
   set +e
-  mvn -e -V -B install -f testsuite
+  mvn -e -V -B clean install -f testsuite
+  mvn -e -V -B test -f testsuite -Pkafka-2_4_1
+  mvn -e -V -B test -f testsuite -Pkafka-2_4_0
```
scholzj, line 34: Why have you chosen these two versions for Travis? Why not for example 2.4.1 + 2.5.0?

mstruk: 2.5.0 is already active without any profile set. Thus, the three mvn invocations actually run the testsuite with 2.5.0, 2.4.1, and 2.4.0.

PR #66, opened 2020-06-30
Token expiry and invalidation improvements

mstruk:
- Documented re-authentication support
- Added immediate JWKS keys refresh upon detection of an unknown `kid` to facilitate recovery when signing keys are aggressively rotated 
- Added session invalidation on access token expiry to `KeycloakRBACAuthorizer`
- Added periodic refresh of grants for active sessions to `KeycloakRBACAuthorizer` for permissions changes to take effect mid-session
- Added a dedicated authorizer - `OAuthSessionAuthorizer` - for session invalidation on access token expiry when re-authentication is not used, and `KeycloakRBACAuthorizer` is not used 
- Deprecated the `JwtKafkaPrincipalBuilder` in favor of the new `io.strimzi.kafka.oauth.server.OAuthKafkaPrincipalBuilder`
- Improved exception handling and throwing during token validation (authentication), so clients can properly handle expired tokens 
- Improved logging inside validators, authorizers, and service components
- Refactored internals to instantiate service components (singletons) where that is called for
- Fixed deprecation warnings in `KeycloakRBACAuthorizer`


oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipalBuilder.java
```diff
@@ -0,0 +1,134 @@
...
+                }
+                sslPrincipalMapper = m.invoke(null, sslPrincipalMappingRules);
+
+            } catch (NoSuchMethodException ex) {
+                Method m = clazz.getMethod("fromRules", String.class);
+                if (sslPrincipalMappingRules == null) {
+                    sslPrincipalMappingRules = "DEFAULT";
+                }
+                sslPrincipalMapper = m.invoke(null, sslPrincipalMappingRules);
+            }
```
tombentley, line 98: I don't understand why the reflection is necessary here. It's a public class and method. It's not in one of the public packages though, which _is_ problematic from a maintainability PoV, but still doesn't necessitate the reflection.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipalBuilder.java
```diff
@@ -0,0 +1,134 @@
...
+
+            // Hack setting sslPrincipalMapper to DefaultKafkaPrincipalBuilder
+            // An alternative would be to copy paste the complete DefaultKafkaPrincipalBuilder implementation
+            // into this class and extend it
+
+            SET_PRINCIPAL_MAPPER.invoke(this, sslPrincipalMapper);
```
tombentley, line 104: This bit is horrible though. Can't you just instantiate your own `SslPrincipalMapper` using the `configs` and not subclass `DefaultKafkaPrincipalBuilder` at all. This looks like it could just break with a new upstream Kafka which made changes in this area.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthSessionAuthorizer.java
```diff
@@ -0,0 +1,159 @@
...
+
+        if (!(principal instanceof OAuthKafkaPrincipal)) {
+            // If user wasn't authenticated over OAuth, there's nothing for us to check
+            if (delegate != null) {
+                return delegate.authorize(session, operation, resource);
+            } else {
+                if (GRANT_LOG.isDebugEnabled()) {
+                    GRANT_LOG.debug("Authorization GRANTED - no access token: " + session.principal() + ", operation: " + operation + ", resource: " + resource);
+                }
+                return true;
```
tombentley, line 94: I can see that you want the delegate to be optional, which sort-of makes sense, but I can also see that the user is one typo away from a default allow, which makes me uncomfortable.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthSessionAuthorizer.java
```diff
@@ -0,0 +1,159 @@
...
+                GRANT_LOG.debug("Authorization GRANTED - access token still valid: " + session.principal() + ", operation: " + operation + ", resource: " + resource + ", token: " + mask(token.value()));
+            }
+            return true;
+        }
+
+        return delegate.authorize(session, operation, resource);
+    }
+
+    private boolean denyIfTokenInvalid(BearerTokenWithPayload token) {
+        if (token.lifetimeMs() <= System.currentTimeMillis()) {
```
tombentley, line 115: Why is it called `lifetimeMs` (which sounds like a duration) rather than `expiryTime` (or `expiryTimeOffset` or something if you want to make the datum clearer).

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthSessionAuthorizer.java
mstruk: Good point. We could add another option e.g. strimzi.authorizer.deny.if.no.delegate=false.
If option is not specified this way we would return false when no delegate is configured.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthSessionAuthorizer.java
mstruk: That's the question for Kafka developers: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/security/oauthbearer/OAuthBearerToken.java#L76-L86

RELEASE_NOTES.md
```diff
@@ -1,6 +1,103 @@
...
+
+* `strimzi.authorization.refresh.grants.period.seconds`
+
+  The time between two grants refresh job runs. 
+  The default value is 60 seconds. If this value is set to 0 or less, refreshing of grants is turned off.
+
+* `strimzi.authorization.refresh.grants.pool.size`
+
+  The number of threads that can fetch grants in parallel.
+  The default value is 5.
```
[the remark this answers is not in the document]
mstruk: No idea. Just seems like a sensible default. It is sized for small deployments for sure. If one thread gets stalled you don't want everything to stop and wait, so it should be more than one thread. On the other hand, there's no point in having too many idle threads just eating memory, inflating thread dumps, while never doing anything.
A quick calculation - let's say a single request takes 50ms (for all I know this could in reality be x10 as fast or x10 as slow). But let's say it's 50ms, then one thread can do 20 requests a second, and 5 threads would do a 100 - you can refresh grants for 100 active sessions with 5 threads in one second.

RELEASE_NOTES.md
```diff
@@ -1,6 +1,103 @@
...
+  
+#### Introduced fast JWKS keys refresh
+
+When using fast local signature validation using JWKS endpoint keys, if signing keys are suddenly revoked at the authorization server, it takes a while for the Kafka broker to be aware of this change.
+Until then, the Kafka broker keeps succesfully authorizing new connections with access tokens signed using old keys.
+At the same time it rejects newly issued access tokens that are properly signed with the new signing keys which Kafka broker doesn't yet know about.
+In order to shorten this mismatch period as much as possible, the broker will now trigger JWKS keys refresh as soon as it detects a new signing key.
+And it will keep trying if not successful using a so called exponential back-off, where after the unsuccessful attempt it waits a second, then two, then 4, 8, 16, 32, ...
```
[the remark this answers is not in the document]
mstruk: The way it's used in JWTSignatureValidator It is bounded via cutoffIntervalSeconds to JWKS refresh seconds period. It wouldn't make much sense to schedule the next refresh one month from now :)

examples/producer/src/main/java/io/strimzi/examples/producer/ExampleConcurrentProducer.java
```diff
@@ -0,0 +1,194 @@
...
+        } finally {
+            producer.close();
+        }
+    }
+
+    @SuppressWarnings("deprecation")
+    private static boolean isAccessTokenJwt(Config config) {
+        String legacy = config.getValue(Config.OAUTH_TOKENS_NOT_JWT);
+        if (legacy != null) {
+            System.out.println("[WARN] OAUTH_TOKENS_NOT_JWT is deprecated. Use OAUTH_ACCESS_TOKEN_IS_JWT (with reverse meaning) instead.");
```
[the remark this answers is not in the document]
mstruk: The idea is that it's a CLI tool, and some output is directed to the user regardless of whether logging is enabled or not. This one detects the use of a deprecated option (passed as ENV var) and this way we make sure the user can see the message.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthSessionAuthorizer.java
tombentley: Yeah, that would work. Maybe we should also WARN about configs beginning with `strimzi.authorizer` which this class doesn't understand (or are you expecting the delegate might also consume `strimzi.authorizer.*` configs)?

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthSessionAuthorizer.java
tombentley: That's a fair point!

RELEASE_NOTES.md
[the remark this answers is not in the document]
tombentley: OK great

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/Sessions.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public List<SessionFuture> executeTask(ExecutorService executor, Predicate<BearerTokenWithPayload> filter,
+                                           Consumer<BearerTokenWithPayload> task) {
+        cleanupExpired();
+
+        // In order to prevent the possible ConcurrentModificationException in the middle of using an iterator
+        // we first make a local copy, then iterate over the copy
+        ArrayList<BearerTokenWithPayload> values;
+        synchronized (activeSessions) {
```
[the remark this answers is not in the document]
mstruk: I wanted to err on the side of caution here without looking at the implementation details of SynchronizedCollection. since `new ArrayList(activeSessions.keySet())` first gets the keySet in one call, and then uses the keyset to make a new ArrayList in another call which, without looking at the code, could use some internal magic that would result in two separate locks on the underlying set.
But looking at the code of SynchronizedMap, the keySet is a singleton wrapped into SynchronizedSet, and toArray() method is synchronized on the Set, and thus it looks like it should work properly without synchronizing on the map.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/Sessions.java
```diff
@@ -0,0 +1,75 @@
...
+                results.add(current);
+            }
+        }
+
+        return results;
+    }
+
+    public void cleanupExpired() {
+        // In order to prevent the possible ConcurrentModificationException in the middle of using an iterator
+        // we first make a local copy, then iterate over the copy
```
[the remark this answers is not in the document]
mstruk: Right, but then there's Sessions.remove() method which can remove from the sessionMap from another thread so iterating the keySet can throw ConcurrentModificationException which we don't want. That's why I make a copy, and iterate over the copy.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/BackOffTaskScheduler.java
```diff
@@ -0,0 +1,151 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * This scheduler adds
+ */
+public class BackOffTaskScheduler {
```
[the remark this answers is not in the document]
mstruk: Good point.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/BackOffTaskScheduler.java
```diff
@@ -0,0 +1,151 @@
...
+                // Limit the delay to maxIntervalSeconds
+                // That makes it possible to grow delay exponentially to some maximum, and then keep it constant
+                if (maxIntervalSeconds > 0 && delay > maxIntervalSeconds) {
+                    delay = maxIntervalSeconds;
+                }
+
+                // Only reschedule if the next run would happen within cutoffIntervalSeconds
+                // If there is another periodic job then we may want to stop re-scheduling this one
+                // once the other job's period is reached.
+                if (cutoffIntervalSeconds <= 0 || delay < cutoffIntervalSeconds) {
```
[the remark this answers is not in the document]
mstruk: Yes, in this case the execution ended with exception and we are rescheduling thus, we're still 'holding' the task execution. It will be released when run() method completes without exception or when we give up rescheduling.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthSessionAuthorizer.java
mstruk: I'd say let's not overdo this. Users should read the docs, and the docs should be obvious about options for specific authorizer.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipalBuilder.java
mstruk: Maybe, but I'd leave this for another PR.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
```diff
@@ -267,89 +322,124 @@ static HostnameVerifier createHostnameVerifier(Config config) {
...
-                    if (grantedScopes.isGranted(operation.name())) {
-                        if (GRANT_LOG.isDebugEnabled()) {
-                            GRANT_LOG.debug("Authorization GRANTED - cluster: " + clusterName + ", user: " + session.principal() + ", operation: " + operation +
-                                    ", resource: " + resource + "\nGranted scopes for resource (" + resourceSpec + "): " + grantedScopes);
+            //
+            // Iterate authorization rules and try to find a match
+            //
+
+            if (grants != null) {
+                Iterator<JsonNode> it = grants.iterator();
```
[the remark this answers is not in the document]
mstruk: But of course!

examples/producer/src/main/java/io/strimzi/examples/producer/ExampleConcurrentProducer.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.List;
+import java.util.Properties;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+
+public class ExampleConcurrentProducer {
+
+    private static Logger log = LoggerFactory.getLogger(ExampleConcurrentProducer.class);
+
+    public static void main(String[] args) {
```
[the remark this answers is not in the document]
mstruk: Refactored it a little bit.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/BackOffTaskScheduler.java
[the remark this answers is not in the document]
mstruk: However, for now this doesn't seem to be an issue - so maybe we leave this one for some other time maybe, when there is a test we're having a trouble with.

README.md
```diff
@@ -342,6 +350,57 @@ Also specify the principal corresponding to the client account identified by `oa
 
 This is not a full set of available `oauth.*` properties. All the `oauth.*` properties described in the next chapter about [configuring the Kafka clients](#configuring-the-kafka-client) also apply to configuring the client side of inter-broker communication. 
 
+#### Enabling the re-authentication
+
+Access tokens expire after some time. The token validation for the purpose of session authentication is only performed immediately after the new connection from the client is established. 
+If using SimpleACLAuthorizer or no authorizer at all, then there is no further need for the access token after the validation, and by default the expiry of the token will not result in session closure or denial of access within the existing session.
```
scholzj, line 356: MAybe this would be more clear for regular users?

```suggestion
If using SimpleACLAuthorizer or no authorizer at all, then there is no further need for the access token after the authentication, and by default the expiry of the token will not result in session closure or denial of access within the existing session.
```

README.md
```diff
@@ -342,6 +350,57 @@ Also specify the principal corresponding to the client account identified by `oa
...
+When a client sends the new access token, validation is performed on the broker as if a new connection was established.
+
+The mechanism has to be explicitly enabled on the Kafka broker by using the `connections.max.reauth.ms` property in `server.properties`:
+
+    connections.max.reauth.ms=3600000
+
+In this example the maximum time until next re-authentication is set to one hour.
+If the access token expires sooner than that, the re-authentication will be triggered sooner.
+
+The option can be specified per-listener. For example if you have a listener called `CLIENTS`, you can specify:
```
scholzj, line 368: Is it disabled by default? Or does it have some default value?

README.md
```diff
@@ -342,6 +350,57 @@ Also specify the principal corresponding to the client account identified by `oa
...
+    listener.name.clients.oauthbearer.connections.max.reauth.ms=3600000
+
+#### Enforcing the session timeout 
+
+If re-authentication is enabled, the session timeout is enforced as the expiry time of the access token. 
+By using re-authentication the multiple 'lightweight' sessions can follow one another over the same network connection for as long as the connection isn't closed or interrupted due to processes restarting or due to network issues. 
+
+If for some reason you can't enable re-authentication or don't want to use it, and if you want to invalidate the session when access token expires, but aren't using `KeycloakRBACAuthorizer`, which does this automatically (since version 0.6.0 of this library), you can use the `OAuthSessionAuthorizer` to enforce token expiry mid-session.
+
+`OAuthSessionAuthorizer` works by checking the access token expiry on every operation performed, and denies all access after the token has expired.
```
scholzj, line 379: Will we add support for this to Strimzi as well?

mstruk:
> Some nits, but LGTM.
> 
> I think you build there quite a nice documentation. I wonder if we should change it from regular README.md file to some better docs format and have it published on the Strimzi website. WDYT? (that would be beyond this PR of course)

Sure, but I'd prefer that this README either remains the source of truth or that another document that is the source of truth is part of this repo, and any web page is rendered from it. I'd also prefer that it remains readable through GitHub so that limits us to Markdown and AsciiDoc.

README.md
mstruk: It is actually disabled by default, which I didn't expect, but the reason is that it relies on additions to Kafka wire protocol introduced in Kafka 2.2, and the clients that don't understand it could break.

scholzj:
> Sure, but I'd prefer that this README either remains the source of truth or that another document that is the source of truth is part of this repo, and any web page is rendered from it. I'd also prefer that it remains readable through GitHub so that limits us to Markdown and AsciiDoc.

Yeah ... it should definitely stay in this repo and be pushed from here. We do the same for Bridge for example. So that is not a problem. I guess Asciidoc might be better since we use it elsewhere and it will give the rendered HTML document the same look & feel. But we already abandoned the modular docs - so it could be simply one Asciidoc file for a start. Anyway, as I said, not in scope for this PR. But it would allow to have this more visible also on the website which IMHO the work you put into this deserves.

README.md
mstruk: This one no. I don't see the need. In order to not break compatibility with clients relying on current behaviour where session never expires due to re-authentication being disabled by default, I'll introduce the oauth authentication setting to turn on re-authentication. Once re-authentication is enabled, there is no need for `OAuthSessionAuthorizer`.

README.md
scholzj: Maybe you can add a note about it here? It wasn't completely clear to me.

mstruk: I added a little more text.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/BackOffTaskScheduler.java
```diff
@@ -0,0 +1,186 @@
...
+            return true;
+        }
+        return false;
+    }
+
+    /**
+     * Update the time of last execution attempt to current time.
+     * This is handy when multiple schedulers work the same task on a single thread.
+     */
+    public void updateLastExecutionTime() {
```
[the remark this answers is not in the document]
mstruk: But then I found a way where there's no need for this method, so it's going away.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
```diff
@@ -31,11 +31,10 @@
 import java.security.Provider;
 import java.security.PublicKey;
 import java.security.Security;
+import java.util.Collections;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ThreadFactory;
```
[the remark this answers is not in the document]
mstruk: Added more javadoc.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/BackOffTaskScheduler.java
```diff
@@ -34,50 +44,50 @@
...
-    private int maxIntervalSeconds;
-    private int cutoffIntervalSeconds;
+    private final int minPauseSeconds;
+    private final int cutoffIntervalSeconds;
 
     // Only one task is scheduled at a time
-    private AtomicBoolean taskScheduled = new AtomicBoolean(false);
+    private final AtomicBoolean taskScheduled = new AtomicBoolean(false);
 
     private long lastExecutionAttempt;
```
tombentley, line 53: Should this be `volatile`? It's written when running that task and read in `scheduleTask()`, which is called from `io.strimzi.kafka.oauth.validator.JWTSignatureValidator#validate` which I think is running in a different thread.

mstruk:
Please, don't merge yet, I found some more bugs and have a few more updates to push.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/BackOffTaskScheduler.java
mstruk: I don't think it's a problem, since scheduleTask() method isn't cycling in a loop it needs to break out of based on shared variable value. Even if I'm wrong about that and on super rare occasion it would happen that scheduleTask() would not see the latest updated lastExecutionAttempt, then the pause before the next execution would just be shorter than otherwise.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ValidatorKey.java
```diff
@@ -0,0 +1,256 @@
...
+    String validIssuerUri;
+    String usernameClaim;
+    String fallbackUsernameClaim;
+    String fallbackUsernamePrefix;
+
+    String sslTruststore;
+    String sslStorePassword;
+    String sslStoreType;
+    String sslRandom;
+    boolean hasHostnameVerifier;
```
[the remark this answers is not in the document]
mstruk: I removed the builders, and made these private final.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ValidatorKey.java
```diff
@@ -0,0 +1,256 @@
...
+
+    public static class JwtValidatorKey extends ValidatorKey {
+
+        String jwksEndpointUri;
+        int jwksRefreshSeconds;
+        int jwksExpirySeconds;
+        int jwksRefreshMinPauseSeconds;
+        boolean checkAccessTokenType;
+        boolean enableBouncy;
+        int bouncyPosition;
```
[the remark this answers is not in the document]
mstruk: I removed the builders, and made these private final.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/BackOffTaskScheduler.java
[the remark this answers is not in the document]
mstruk: I added CurrentTimeProvider, and a test that uses it.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
```diff
@@ -103,34 +107,96 @@ public void configure(Map<String, ?> configs, String saslMechanism, List<AppConf
...
+                    .sslRandom(sslRnd)
+                    .hasHostnameVerifier(verifier != null)
+                    .jwksEndpointUri(jwksUri)
+                    .jwksRefreshSeconds(jwksRefreshSeconds)
+                    .jwksRefreshMinPauseSeconds(jwksMinPauseSeconds)
+                    .jwksExpirySeconds(jwksExpirySeconds)
+                    .checkAccessTokenType(checkTokenType)
+                    .enableBouncy(enableBouncy)
+                    .bouncyPosition(bouncyPosition)
+                    .build();
```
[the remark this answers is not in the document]
mstruk: I removed the builders, and made the constructors that take "20 arguments".

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ValidatorKey.java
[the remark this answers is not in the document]
tombentley: Just to be clear, my problem was with having builders _and_ mutable fields. If the fields are mutable then there's no need really for the builder. If the fields are immutable then using a builder to avoid a ctor with a stupid number of parameters is OK. Or you just have the constructor with the stupid number of parameters. Since the class is only ever instantiated once then having the constructor is acceptable (to me at least). One benefit of the ctor is it forces the caller to provide all the arguments. You can do that with the builder too, of course, but it becomes a runtime error and it make the builder more complicated.

mstruk:
@scholzj I think this is good to merge.

PR #68, opened 2020-08-11
Support for Kafka 2.6 and some scripts and logging fixes

mstruk:


.travis/build.sh
```diff
@@ -32,7 +32,8 @@ if [ ${JAVA_MAJOR_VERSION} -eq 1 ] ; then
   docker pull oryd/hydra:v1.0.0
   mvn test-compile spotbugs:check -e -V -B -f testsuite
   set +e
-  mvn -e -V -B clean install -f testsuite
+  mvn -e -V -B clean install -f testsuite -Pkafka-2_6_0
+  mvn -e -V -B test -f testsuite -Pkafka-2_5_0
   mvn -e -V -B test -f testsuite -Pkafka-2_4_1
   mvn -e -V -B test -f testsuite -Pkafka-2_4_0
```
ppatierno, line 38: we removed these from the operator. should we still support them here?

testsuite/README.md
```diff
@@ -49,6 +49,7 @@ There are several profiles available to test with a specific version of Kafka im
 - kafka-2_4_0
 - kafka-2_4_1
```
ppatierno, line 50: same questions as above.

.travis/build.sh
mstruk: It does consume a bit more resources when running CI, but for the time being I see no reason to not keep supporting the older versions. People may also be using strimzi-kafka-oauth independently of strimzi-kafka-operator for their own cloud or stand-alone deployments. It's good to detect if something we do breaks compatibility with older versions so we can consider it and mention it in release notes once it happens.

ppatierno: It makes sense, thanks.

PR #73, opened 2020-09-24
Prepare for 0.6.1 release

mstruk:
I suggest replacing <version> in all poms. I've encountered situations where the tests or examples were using 1.0.0-SNAPSHOT rather than the just built 0.6.x tag, resulting in non-reproducable build, non-working examples and similar.


mstruk:
Maybe too late for 0.6.1 release but I suggest we used this approach for future releases.

scholzj (approved):

scholzj:
Too late for 0.6.1. But I will merge it to have it if we do another patch release.

PR #77, opened 2020-11-15
OAuth over SASL_PLAIN

mstruk:
OAuth over PLAIN

An implementation of using SASL_PLAIN to authenticate with a clientId and a secret or with an access token.
See README.md for instructions on how to set it up.


examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
```diff
@@ -114,11 +114,17 @@ private static Properties buildProducerConfig() {
...
+
+        //p.setProperty("security.protocol", "SASL_PLAINTEXT");
+        //p.setProperty("sasl.mechanism", "OAUTHBEARER");
+        //p.setProperty("sasl.jaas.config", "org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required ;");
+        //p.setProperty("sasl.login.callback.handler.class", "io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler");
```
scholzj, line 123: Is it intentional to have single example with both OAUTHBEARER and PLAIN in one like this? If yes, it might be nice to add some explanatory comments here in the code to explain what and when should be un-commented and what should be commented.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/LogUtil.java
```diff
@@ -35,4 +35,20 @@ public static String mask(String input) {
...
         return input.substring(0, 4) + "**" + input.substring(len - 4, len);
     }
+
+    /**
+     * Collect all exception causes into a single string
+     *
+     * @param e The exception
+     * @return The message with all the causes
+     */
+    public static String getCauseMessage(Throwable e) {
```
scholzj, line 45: Should it be maybe named in plural? `getCauseMessages`? Or maybe even `getAllCauseMessages` to make this a bit more self-explaining?

oauth-server-plain/src/main/java/io/strimzi/kafka/oauth/server/plain/JaasServerOauthOverPlainValidatorCallbackHandler.java
```diff
@@ -0,0 +1,124 @@
...
+            throw new SaslAuthenticationException("Authentication failed: " + getCauseMessage(e), e);
+        }
+    }
+
+    private void authenticate(String username, String password) throws UnsupportedCallbackException, IOException {
+        // if username equals 'access-token' we treat the password as an access token
+        // otherwise, we treat username as clientId, password as client secret and perform client credential auth
+        // to get the access token
+        final String accessToken;
+        if ("access-token".equals(username)) {
```
scholzj, line 106: I wonder if we should give this name a bit more special feeling. for example `$accessToken` This fits with what for example Microsoft does on EventHubs. I think it makes it a bit more clear that it is some special value with special meaning when reading the different configuration files using it.

scholzj (commented):
Some nits, but LGTM otherwise.

oauth-server-plain/src/main/java/io/strimzi/kafka/oauth/server/plain/JaasServerOauthOverPlainValidatorCallbackHandler.java
mstruk: Makes sense, but for this specific suggestion of `$accessToken` my concern is that it would cause problems in the CLI use-cases. E.g.:

    kafkacat -b my-cluster-kafka-bootstrap:9092 -X security.protocol=SASL_PLAINTEXT -X sasl.mechanism=PLAIN -X sasl.username=$accessToken -X sasl.password="$ACCESS_TOKEN" -t a_topic -P

This will result in empty username, and authentication will fail. For that reason I would not use $ in the reserved word.
But maybe we could use @ as in `@accessToken`.

scholzj: MS is using `$` ... that was why I suggested it. But you have a good point, although I'm not sure how often will people really use it from the command line. 

@strimzi/maintainers WDYT?.

examples/producer/src/main/java/io/strimzi/examples/producer/ExampleProducer.java
mstruk: Good catch, this was not intended this way.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/LogUtil.java
mstruk: Makes sense.

oauth-server-plain/src/main/java/io/strimzi/kafka/oauth/server/plain/JaasServerOauthOverPlainValidatorCallbackHandler.java
```diff
@@ -0,0 +1,124 @@
...
+import javax.security.auth.login.AppConfigurationEntry;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.List;
+import java.util.Map;
+
+import static io.strimzi.kafka.oauth.common.LogUtil.getAllCauseMessages;
+
+public class JaasServerOauthOverPlainValidatorCallbackHandler extends JaasServerOauthValidatorCallbackHandler {
```
tombentley, line 31: Javadoc

oauth-server-plain/src/main/java/io/strimzi/kafka/oauth/server/plain/JaasServerOauthOverPlainValidatorCallbackHandler.java
tombentley: I don't have a strong opinion. If it's open to interpolation by the shell it can be enclosed in `'`.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipal.java
```diff
@@ -35,6 +37,19 @@ public BearerTokenWithPayload getJwt() {
...
 
+    public static OAuthKafkaPrincipal popCurrentPrincipal() {
+        OAuthKafkaPrincipal current = currentTL.get();
+        currentTL.remove();
+        return current;
+    }
+
+    public static void pushCurrentPrincipal(OAuthKafkaPrincipal principal) {
+        if (currentTL.get() != null) {
+            throw new RuntimeException("Internal error - current principal already set");
```
tombentley, line 48: If these don't support stack-like push and pop semantics then are the method names really the right ones?

README.md
```diff
@@ -358,7 +383,7 @@ If using SimpleACLAuthorizer or no authorizer at all, then there is no further n
 Since Kafka version 2.2 the Kafka brokers support a [re-authentication mechanism](https://cwiki.apache.org/confluence/display/KAFKA/KIP-368%3A+Allow+SASL+Connections+to+Periodically+Re-Authenticate) allowing clients to update the token mid-session, without having to drop and re-establish the connection. 
 When a client sends the new access token, validation is performed on the broker as if a new connection was established.
 
-The mechanism is disabled by default in order to prevent the old clients that don't understand the new Kafka wire protocol from breaking.
+The re-authentication only works with SASL_OAUTHBEARER mechanism, and is disabled by default in order to prevent the old clients that don't understand the new Kafka wire protocol from breaking.
```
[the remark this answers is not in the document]
mstruk: It refers to the re-authentication mechanism which only exists to support SASL mechanisms that are token based, and require token refresh in order for the client to be able to keep using the current connection. When you authenticate with SASL_OAUTHBEARER the Kafka broker will go through a dialog with the client when the token expires. It won't ever enter that dialog when you authenticate with SASL_PLAIN and the token expires. In the latter case any additional interaction will simply result in the `AuthorizationException`.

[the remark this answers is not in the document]
mstruk: The part about being disabled by default refers to the fact that unless you specify `maxSecondsWithoutReauthentication` config option the re-authentication won't even be activated for SASL_OAUTHBEARER.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipal.java
mstruk: Fair point. Will rename.

oauth-server-plain/src/main/java/io/strimzi/kafka/oauth/server/plain/JaasServerOauthOverPlainValidatorCallbackHandler.java
mstruk: Ok, so I guess '$accessToken' seems more natural to most people, and they would understand this as setting a kind of server side variable rather than refer to the client side environment variable. And escaping it in the shell is simple.

README.md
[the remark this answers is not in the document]
mstruk: I reworded it to make it more clear (hopefully).

oauth-server-plain/src/main/java/io/strimzi/kafka/oauth/server/plain/JaasServerOauthOverPlainValidatorCallbackHandler.java
mstruk: Added Javadoc, and to [JaasServerOauthValidatorCallbackHandler](https://github.com/strimzi/strimzi-kafka-oauth/pull/77/files#diff-0af30cbf06d7deee32bafafb11afc8d0f5ac5242878a6e7746f4f315ca94ea11) as well.

tombentley (approved):

scholzj (approved):

scholzj:
@mstruk Can this be merged? Should we do 0.7.0-rc1 after merging this or will there be some more work needed for 0.7.0?

mstruk:
@scholzj It is ready for merge.

I'm working on another PR for 0.7.0 - a functionality to allow enforcing the audience - a mechanism where the token is specifically issued for use by a limited set of resource servers. That makes it possible for Kafka broker to reject the tokens not issued for it during authentication already. While KeycloakRBACAuthorizer performs all the necessary checks during each authorization, one might want to turn off authorization (maybe during development) and still limit the access by only allowing some accounts to authenticate.

scholzj:
@mstruk Ok, thanks. I will wait with the RC then.

PR #78, opened 2020-11-30
Audience checking

mstruk:
Support for audience checking.

When enabled the Kafka broker rejects during authentication any access token that does not contain the Kafka broker's OAuth client ID in`aud` claim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ Specify the following `oauth.*` properties:
 Some authorization servers don't provide the `iss` claim. In that case you would not set `oauth.valid.issuer.uri`, and you would explicitly turn off issuer checking by setting the following option to `false`:
 - `oauth.check.issuer` (e.g. "false")
 
-You can enforce audience checking, which is an OAuth2 mechanism to prevent successful authentication with tokens that are not explicitly issued for use by your resource server.
+You can enforce audience checking, which is an OAuth2 mechanism to prevent successful authentication with tokens unless they are explicitly issued for use by your resource server.
 The authorization server adds the allowed resource servers' `client IDs` into the `aud` claim of such tokens.
 
 Set the following option to `true` to enforce audience checking:
@@ -332,7 +332,7 @@ Introspection endpoint should be protected. The `oauth.client.id` and `oauth.cli
 Some authorization servers don't provide the `iss` claim. In that case you would not set `oauth.valid.issuer.uri`, and you would explicitly turn off issuer checking by setting the following option to `false`:
 - `oauth.check.issuer` (e.g.: "false")
 
-You can enforce audience checking, which is an OAuth2 mechanism to prevent successful authentication with tokens that are explicitly issued for use by your resource server.
+You can enforce audience checking, which is an OAuth2 mechanism to prevent successful authentication with tokens unless they are explicitly issued for use by your resource server.
 The authorization server adds the allowed resource servers' `client IDs` into the `aud` claim of such tokens.
 
 Set the following option to `true` to enforce audience checking:
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ The next thing to do is to enable SASL_OAUTHBEARER mechanism:
 
     sasl.enabled.mechanisms=OAUTHBEARER
 
-Since version 0.7.0 there is also support for so called 'OAuth over PLAIN' which allows using the SASL_PLAIN mechanism to authenticate with an OAuth access token or with a clientId and a secret.
+Since version 0.7.0 there is also support for so called 'OAuth over PLAIN' which allows using the SASL_PLAIN mechanism to authenticate with an OAuth access token or with a client ID and a secret.
 In order to use that you have to enable the SASL_PLAIN mechanism as well (you can enable one or the other or both):
 
     sasl.enabled.mechanisms=OAUTHBEARER,PLAIN
@@ -265,6 +265,14 @@ Specify the following `oauth.*` properties:
 Some authorization servers don't provide the `iss` claim. In that case you would not set `oauth.valid.issuer.uri`, and you would explicitly turn off issuer checking by setting the following option to `false`:
 - `oauth.check.issuer` (e.g. "false")
 
+You can enforce audience check, which is an OAuth2 mechanism to limit access to tokens that are explicitly issued for use by your protected server.
+The authorization server adds the allowed resource servers' `client ids` into `aud` claim of such tokens.
+Set the following option to `true` to enforce audience check (it is not enabled by default). 
+- `oauth.check.audience` (e.g. "true")
+
+When audience check is enabled the `oauth.client.id` has to be configured:
+- `oauth.client.id` (e.g.: "kafka" - this is the OAuth2 client configuration id for Kafka Broker)
+
 JWT tokens contain unique user identification in `sub` claim. However, this is often a long number or a UUID, but we usually prefer to use human readable usernames, which may also be present in JWT token.
 Use `oauth.username.claim` to map the claim (attribute) where the value you want to use as user id is stored:
 - `oauth.username.claim` (e.g.: "preferred_username")
@@ -311,13 +319,18 @@ This will result in Kafka Broker making a request to authorization server every 
 Specify the following `oauth.*` properties:
 - `oauth.introspection.endpoint.uri` (e.g.: "https://localhost:8443/auth/realms/demo/protocol/openid-connect/token/introspect")
 - `oauth.valid.issuer.uri` (e.g.: "https://localhost:8443/auth/realms/demo" - only access tokens issued by this issuer will be accepted)
-- `oauth.client.id` (e.g.: "kafka" - this is the client configuration id for Kafka Broker)
+- `oauth.client.id` (e.g.: "kafka" - this is the OAuth2 client configuration id for Kafka Broker)
 - `oauth.client.secret` (e.g.: "kafka-secret")
  
 Introspection endpoint should be protected. The `oauth.client.id` and `oauth.client.secret` specify Kafka Broker credentials for authenticating to access the introspection endpoint. 
 
 Some authorization servers don't provide the `iss` claim. In that case you would not set `oauth.valid.issuer.uri`, and you would explicitly turn off issuer checking by setting the following option to `false`:
 - `oauth.check.issuer` (e.g.: "false")
+
+You can enforce audience check, which is an OAuth2 mechanism to limit access to tokens that are explicitly issued for use by your protected server.
+The authorization server adds the allowed resource servers' `client IDs` into `aud` claim of such tokens.
+Set the following option to `true` to enforce audience check (it is not enabled by default). 
+- `oauth.check.audience` (e.g. "true")
 
 By default, if the Introspection Endpoint response contains `token_type` claim, there is no checking performed on it.
 Some authorization servers use a non-standard `token_type` value. To give the most flexibility, you can specify the valid `token_type` for your authorization server:
@@ -362,7 +375,7 @@ When you have a DEBUG logging configured for the `io.strimzi` category you may n
 
 ##### Configuring the client side of inter-broker communication
 
-All the Kafka Brokers in the cluster should be configured with the same clientId and clientSecret, and the corresponding user should be added to `super.users` since inter-broker client requires super-user permissions.
+All the Kafka Brokers in the cluster should be configured with the same client ID and secret, and the corresponding user should be added to `super.users` since inter-broker client requires super-user permissions.
 
 Specify the following `oauth.*` properties:
 - `oauth.token.endpoint.uri` (e.g.: "https://localhost:8443/auth/realms/demo/protocol/openid-connect/token")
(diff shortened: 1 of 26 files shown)


tombentley:
@mstruk any links to documentation about what "audience check" _is_?

mstruk:
> @mstruk any links to documentation about what "audience check" _is_?

I don't know about a simple nice explanation.
It's mentioned in several places in the [OpenID Connect spec (extension of OAuth2)](https://openid.net/specs/openid-connect-core-1_0.html#IDToken).

scholzj:
> I don't know about a simple nice explanation.
It's mentioned in several places in the OpenID Connect spec (extension of OAuth2).

It will anyway need to be documented. So I guess if there is no reasonable explanation what this does we (well, you) would need to write it.

README.md
```diff
@@ -265,6 +265,14 @@ Specify the following `oauth.*` properties:
 Some authorization servers don't provide the `iss` claim. In that case you would not set `oauth.valid.issuer.uri`, and you would explicitly turn off issuer checking by setting the following option to `false`:
 - `oauth.check.issuer` (e.g. "false")
 
+You can enforce audience check, which is an OAuth2 mechanism to limit access to tokens that are explicitly issued for use by your protected server.
+The authorization server adds the allowed resource servers' `client ids` into `aud` claim of such tokens.
+Set the following option to `true` to enforce audience check (it is not enabled by default). 
+- `oauth.check.audience` (e.g. "true")
```
scholzj, line 271: So, the JWT token contains the `aud` field, right? And the `oauth.check.audience` says that it should be checked. But checked against what? Where do you set the allowed `aud` values?

mstruk: The valid value is only one - the configured `oauth.client.id`. Maybe it's not adequately explained in the README.md?
For example - if client id is configured as 'kafka' which it usually is, then the examples of valid `aud` value are:

* "kafka"
* ["rest-api", "kafka"]

mstruk:
> It will anyway need to be documented. So I guess if there is no reasonable explanation what this does we (well, you) would need to write it.
The additional option is documented in README.md, but maybe not sufficiently. I'll add a sentence or two..

README.md
```diff
@@ -265,6 +265,14 @@ Specify the following `oauth.*` properties:
 Some authorization servers don't provide the `iss` claim. In that case you would not set `oauth.valid.issuer.uri`, and you would explicitly turn off issuer checking by setting the following option to `false`:
 - `oauth.check.issuer` (e.g. "false")
 
+You can enforce audience check, which is an OAuth2 mechanism to limit access to tokens that are explicitly issued for use by your protected server.
```
tombentley: I think "limit access" is a bit confusing. IIUC it's not that there is something which can't _access_ the tokens, it's that the tokens are rejected as invalid because they weren't issued to be used for that Kafka broker (acting in the role of resource server). Maybe "...which is an OAuth2 mechanism to prevent validation of tokens that weren't explicitly issued for use by your protected server." would be better phrasing.

Also, normal English usage would be "You can enforce audience checking".

README.md
```diff
@@ -265,6 +265,14 @@ Specify the following `oauth.*` properties:
...
+The authorization server adds the allowed resource servers' `client ids` into `aud` claim of such tokens.
+Set the following option to `true` to enforce audience check (it is not enabled by default).
```
tombentley:
```suggestion
Set the following option to `true` to enforce audience checking (it is not enabled by default). 
```

README.md
```diff
@@ -265,6 +265,14 @@ Specify the following `oauth.*` properties:
 Some authorization servers don't provide the `iss` claim. In that case you would not set `oauth.valid.issuer.uri`, and you would explicitly turn off issuer checking by setting the following option to `false`:
 - `oauth.check.issuer` (e.g. "false")
 
+You can enforce audience check, which is an OAuth2 mechanism to limit access to tokens that are explicitly issued for use by your protected server.
+The authorization server adds the allowed resource servers' `client ids` into `aud` claim of such tokens.
+Set the following option to `true` to enforce audience check (it is not enabled by default). 
+- `oauth.check.audience` (e.g. "true")
+
+When audience check is enabled the `oauth.client.id` has to be configured:
```
tombentley, line 274:
```suggestion
When audience checking is enabled the `oauth.client.id` has to be configured:
```

README.md
```diff
@@ -265,6 +265,14 @@ Specify the following `oauth.*` properties:
...
+- `oauth.client.id` (e.g.: "kafka" - this is the OAuth2 client configuration id for Kafka Broker)
```
tombentley, line 275:
```suggestion
- `oauth.client.id` (e.g.: "kafka" - this is the OAuth2 client configuration id for the Kafka Broker)
```

README.md
```diff
@@ -311,14 +319,19 @@ This will result in Kafka Broker making a request to authorization server every
 Specify the following `oauth.*` properties:
 - `oauth.introspection.endpoint.uri` (e.g.: "https://localhost:8443/auth/realms/demo/protocol/openid-connect/token/introspect")
 - `oauth.valid.issuer.uri` (e.g.: "https://localhost:8443/auth/realms/demo" - only access tokens issued by this issuer will be accepted)
-- `oauth.client.id` (e.g.: "kafka" - this is the client configuration id for Kafka Broker)
+- `oauth.client.id` (e.g.: "kafka" - this is the OAuth2 client configuration id for Kafka Broker)
```
tombentley, line 327:
```suggestion
- `oauth.client.id` (e.g.: "kafka" - this is the OAuth2 client configuration id for the Kafka Broker)
```

README.md
```diff
@@ -311,14 +319,19 @@ This will result in Kafka Broker making a request to authorization server every
...
 - `oauth.client.secret` (e.g.: "kafka-secret")
  
 Introspection endpoint should be protected. The `oauth.client.id` and `oauth.client.secret` specify Kafka Broker credentials for authenticating to access the introspection endpoint. 
 
 Some authorization servers don't provide the `iss` claim. In that case you would not set `oauth.valid.issuer.uri`, and you would explicitly turn off issuer checking by setting the following option to `false`:
 - `oauth.check.issuer` (e.g.: "false")
 
+You can enforce audience check, which is an OAuth2 mechanism to limit access to tokens that are explicitly issued for use by your protected server.
```
tombentley: Same comment

(also why are we repeating this verbatim , can't we just refer the reader to the previous duscussion?)

mstruk:
> It's mentioned in several places in the [OpenID Connect spec (extension of OAuth2)](https://openid.net/specs/openid-connect-core-1_0.html#IDToken).

[Another link](https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-25#section-4.1.3) (for reference).

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
```diff
@@ -210,7 +213,8 @@ private void performOptionalChecks(JsonNode response) {
 
         if (audience != null) {
             value = response.get("aud");
-            if (value == null || !audience.equals(value.asText())) {
+            List<String> audienceList = value != null ? JSONUtil.asListOfString(value) : Collections.emptyList();
```
tombentley, line 216: When `value` is a JSON string, rather than a JSON array of strings, `JSONUtil.asListOfString` will split it on whitespace, which doesn't seem to be correct according to https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-25 or https://openid.net/specs/openid-connect-core-1_0.html#IDToken. It _might_ be, because "The interpretation of audience values is generally application specific.". 

The JWT document also says that the content of the JSON strings can also be a URI if the String contains a `:`, Is that something we need to think about here?

README.md
mstruk: OTOH when I hear 'enforce audience checking' I specifically think of this functionality being enabled vs. not being enabled. I don't think about the process of validating the token being triggered and failing. Rather I think about the process of validating not being attempted at all.

mstruk: Ah, sorry I misunderstood your last sentence. So with regards to 'limit access', what about saying it like this:

"...which is an OAuth2 mechanism to prevent *authentication* of tokens that weren't explicitly issued for use by your protected server."

README.md
mstruk: It's structured to have 'fast local JWT checking' chapter with all the config options, and a separate 'introspection based validation' chapter. Some options are shared by both. But if you look closely you'll see that there are some nuances and even the part about audience checking is slightly different for 'fast JWT checking' chapter. There's this sentence:

"When audience check is enabled the `oauth.client.id` has to be configured:
- `oauth.client.id` (e.g.: "kafka" - this is the OAuth2 client configuration id for Kafka Broker)" 

Which is not present in 'Introspection based validation' chapter, because `oauth.client.id` is already a required configuration there.

I prefer it this way as it's more user-friendly than having the user jump around and trying to figure out if something fits together with something else.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
mstruk: Good catch about `JSONUtil.asListOfString()` even though I think it behaves slightly different than you described :)

If the value is `"kafka"` then it's a text node, and it will be correctly parsed. 
If the value is `"kafka rest-api"` - that's the application specific handling, and not by spec - it will be correctly parsed under the assumption that the client id can't contain a space character (I have not checked this assumption). 
If the value is `"kafka, rest-api"` it is a text node again and not by the spec - it WON'T be correctly parsed unless there is client id containing a comma and a space (I assume these are not valid characters for client id). I also assume this case would never happen that multiple client ids would be sent this way - as a single string.
If the value is `["kafka", "rest-api"]` or `["kafka"]` it is an array node and will be handled correctly.

We could improve the JSONUtil.asListOfString() where it parses the list of strings sent as a single string to also treat the comma as a separator, not just the space. Assuming that all the contexts where it is used don't allow comma and space as legal characters inside individual string items or that no reasonable person would use these characters in client ID.

The spec indeed says that it could be a url, but I'm unaware of any implementation using that, and if we go testing proper support if url is the value then we're possibly spending time on something that no-one would ever use and that has no real implementation for us to test with.

README.md
mstruk: Or maybe it should be:

"...which is an OAuth2 mechanism to prevent *successful authentication with* tokens that weren't explicitly issued for use by your protected server."

tombentley: SGTM.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
tombentley: The spec seems to be silent about whether the string can contain whitespace or not. I assumed they could, but that such values should not be treated as a list of tokens, but rather a single value containing a string. In any case I'll defer to you on this, I just wanted to point this stuff out.

README.md
```diff
@@ -265,6 +265,19 @@ Specify the following `oauth.*` properties:
 Some authorization servers don't provide the `iss` claim. In that case you would not set `oauth.valid.issuer.uri`, and you would explicitly turn off issuer checking by setting the following option to `false`:
 - `oauth.check.issuer` (e.g. "false")
 
+You can enforce audience checking, which is an OAuth2 mechanism to prevent successful authentication with tokens that are explicitly issued for use by your resource server.
```
tombentley, line 268:
```suggestion
You can enforce audience checking, which is an OAuth2 mechanism to prevent successful authentication with tokens that are not explicitly issued for use by your resource server.
```

README.md
```diff
@@ -311,14 +324,24 @@ This will result in Kafka Broker making a request to authorization server every
...
-- `oauth.client.id` (e.g.: "kafka" - this is the client configuration id for Kafka Broker)
+- `oauth.client.id` (e.g.: "kafka" - this is the OAuth2 client configuration id for the Kafka Broker)
 - `oauth.client.secret` (e.g.: "kafka-secret")
  
 Introspection endpoint should be protected. The `oauth.client.id` and `oauth.client.secret` specify Kafka Broker credentials for authenticating to access the introspection endpoint. 
 
 Some authorization servers don't provide the `iss` claim. In that case you would not set `oauth.valid.issuer.uri`, and you would explicitly turn off issuer checking by setting the following option to `false`:
 - `oauth.check.issuer` (e.g.: "false")
 
+You can enforce audience checking, which is an OAuth2 mechanism to prevent successful authentication with tokens that are explicitly issued for use by your resource server.
```
tombentley, line 335:
```suggestion
You can enforce audience checking, which is an OAuth2 mechanism to prevent successful authentication with tokens that are not explicitly issued for use by your resource server.
```

README.md
```diff
@@ -311,14 +324,24 @@ This will result in Kafka Broker making a request to authorization server every
...
+The authorization server adds the allowed resource servers' `client IDs` into `aud` claim of such tokens.
```
tombentley, line 336:
```suggestion
The authorization server adds the allowed resource servers' `client IDs` into the `aud` claim of such tokens.
```

README.md
```diff
@@ -265,6 +265,19 @@ Specify the following `oauth.*` properties:
 Some authorization servers don't provide the `iss` claim. In that case you would not set `oauth.valid.issuer.uri`, and you would explicitly turn off issuer checking by setting the following option to `false`:
 - `oauth.check.issuer` (e.g. "false")
 
+You can enforce audience checking, which is an OAuth2 mechanism to prevent successful authentication with tokens that are explicitly issued for use by your resource server.
+The authorization server adds the allowed resource servers' `client IDs` into `aud` claim of such tokens.
```
tombentley, line 269:
```suggestion
The authorization server adds the allowed resource servers' `client IDs` into the `aud` claim of such tokens.
```

tombentley (approved):

README.md
mstruk: Good catch, I suggest a slightly different fix to keep consistency with the next sentence.
```suggestion
You can enforce audience checking, which is an OAuth2 mechanism to prevent successful authentication unless the tokens are explicitly issued for use by your resource server.
```

README.md
mstruk: Again ...
```suggestion
You can enforce audience checking, which is an OAuth2 mechanism to prevent successful authentication unless tokens are explicitly issued for use by your resource server.
```

scholzj (approved):

scholzj:
@mstruk Looks good to me ... but you will need to fix the DCO signoff.

scholzj:
And the CI build I guess.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
mstruk: Let's just leave it like this for now.

README.md
mstruk: Slightly reworded to fit with the line that follows:

"You can enforce audience checking, which is an OAuth2 mechanism to prevent *successful authentication with* tokens **unless they are** explicitly issued for use by your resource server."

mstruk:
Testsuite passes, DCO fixed. AFAIC this is good to merge.

PR #83, opened 2020-12-08
Custom claim check

mstruk:
This feature adds `oauth.custom.claim.check` configuration option which accepts a JSONPath-like filter syntax which one can use to express a filter that has to match the JWT token in order for the token to not be rejected during authentication.

The checking can also be performed against the introspection endpoint response.

For more about the syntax see: [JsonPathFilterQuery.java](https://github.com/strimzi/strimzi-kafka-oauth/blob/2c7e4f6e43b2cd5f515635e10ef961e1deeb53c9/oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/JsonPathFilterQuery.java)

This PR is based on custom JSONPath filter syntax implementation. It is superseded by #87 which uses Jayway JsonPath implementation.


oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/ListNode.java
```diff
@@ -0,0 +1,30 @@
...
+class ListNode extends Node {
+    private final ArrayList<Node> items;
+
+    ListNode(ArrayList<Node> items) {
+        this.items = items;
+    }
+
+    @Override
+    public String toString() {
+        return items.stream().map(Object::toString).collect(Collectors.joining(",", "[", "]"));
```
tombentley, line 19: It's this going to produce the same result as simply calling `items.toString`?

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/ListNode.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public boolean contains(Object o) {
+        for (Node n : items) {
```
tombentley, line 23: Can't you use `items.contains(o)`?

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/ParsingContext.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    ParsingContext(char[] buffer, int start, int current) {
+        this.buffer = buffer;
+        this.start = start;
+        this.current = current;
+    }
+
+    @Override
+    protected ParsingContext clone() throws CloneNotSupportedException {
```
tombentley, line 24: Why redeclare the `throws CloneNotSupportedException` when it doesn't throw that?

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/ParsingContext.java
```diff
@@ -0,0 +1,153 @@
...
+        ParsingContext ctx = (ParsingContext) super.clone();
+        ctx.buffer = buffer;
+        ctx.start = start;
+        ctx.current = current;
+        return ctx;
+    }
+
+    public ParsingContext copy() {
```
tombentley, line 32: And if you're not redeclaring the throws in `clone` this method probably isn't needed?

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/ParsingContext.java
```diff
@@ -0,0 +1,153 @@
...
+        } catch (CloneNotSupportedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public void reset() {
+        current = start;
+    }
+
+    public void reset(int offset) {
```
tombentley, line 44: Maybe `resetTo` would be a better name?

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/ParsingContext.java
```diff
@@ -0,0 +1,153 @@
+/*
+ * Copyright 2017-2019, Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.oauth.jsonpath;
+
+class ParsingContext implements Cloneable {
```
tombentley, line 7: Some javadoc on the methods of this class wouldn't go amiss.

oauth-common/src/test/java/io/strimzi/kafka/oauth/jsonpath/JsonPathFilterQueryTest.java
```diff
@@ -0,0 +1,86 @@
...
+            "@.exp > 1000 || 'kafka' in @.aud", "true"
+        };
+
+        String[] errQueries = {
+            "['exp'] > 1000", "attribute path",
+            "@", "at position: 0",
+            "@.custom in null", "right of 'in'"
+        };
+
+        // TODO: Unsupported comparison
```
tombentley, line 59: Are you going to implement this test?

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/JsonPathFilterQuery.java
```diff
@@ -0,0 +1,456 @@
...
+ *   }
+ * </pre>
+ *
+ * Some examples of valid queries are:
+ *
+ * <pre>
+ *   "@.exp &lt; 1000"
+ *   "@.custom == 'custom-value'"
+ *   "@.custom == 'custom-value' and @.exp &gt; 1000"
+ *   "@.custom == 'custom-value' or @.exp &gt;= 1000"
```
tombentley, line 40: What's the precedence of `and` wrt `or`, and can you use brackets for force evaluation order?

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/JsonPathFilterQuery.java
```diff
@@ -0,0 +1,456 @@
...
+        return new JsonPathFilterQuery(query);
+    }
+
+    /**
+     * Match the json objects against the filter query.
+     *
+     * @param jsonObject Jackson DataBind object
+     * @return true if the object matches the filter, false otherwise
+     */
+    public boolean match(JsonNode jsonObject) {
```
tombentley, line 115: `matches` might be slightly nicer?

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/JsonPathFilterQuery.java
```diff
@@ -0,0 +1,456 @@
...
+ *   boolean match = query.match(jsonObject);
+ * </pre>
+ *
+ * Query is parsed in the first line and any errors during parsing result
+ * in {@link JsonPathFilterQueryException}.
+ *
+ * Matching is thread safe. The normal usage pattern is to initialise the JsonPathFilterQuery object once,
+ * and query it many times concurrently against json objects.
+ */
+public class JsonPathFilterQuery {
```
tombentley, line 91: If you're going to write your own parser like this it would be really useful to write down the grammar of the language you're trying to parse. It makes it much easier to read each of the methods to decide whether they're doing the right thing, and it also makes a bit more formal the sorts of input you're accepting.

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/Matcher.java
```diff
@@ -0,0 +1,391 @@
...
+package io.strimzi.kafka.oauth.jsonpath;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.math.BigDecimal;
+import java.util.Iterator;
+
+class Matcher {
```
tombentley, line 14: Javadoc, and say that this is supposed to be thread safe!

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/Matcher.java
```diff
@@ -0,0 +1,391 @@
...
+                } else if (op == OperatorNode.IN) {
+                    eval.update(logical, containedIn(json, predicate));
+                } else if (op == OperatorNode.NIN) {
+                    eval.update(logical, !containedIn(json, predicate));
+                } else if (op == OperatorNode.MATCH_RE) {
+                    throw new RuntimeException("Not implemented");
+                } else if (op == OperatorNode.ANYOF) {
+                    throw new RuntimeException("Not implemented");
+                } else if (op == OperatorNode.NONEOF) {
+                    throw new RuntimeException("Not implemented");
```
tombentley, line 63: If these aren't supported they should be rejected by the parser.

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/ListNode.java
mstruk: So true :D

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/ListNode.java
mstruk: Yup! Will fix it.

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/Matcher.java
mstruk: Still working on these ...

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/ParsingContext.java
mstruk: Good catch. I ended up not needing this method, so I'll just remove it together with copy().

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/ParsingContext.java
mstruk: Well, it's not used anywhere, so I'll remove it.

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/ParsingContext.java
mstruk: Makes sense.

oauth-common/src/test/java/io/strimzi/kafka/oauth/jsonpath/JsonPathFilterQueryTest.java
mstruk: Yes.

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/JsonPathFilterQuery.java
mstruk: Evaluation is left to right, as soon as a false expression is followed by 'and' the evaluation short circuits to false. The brackets would be good yes. I left them out for now. Will see how hard it would be to add them.

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/JsonPathFilterQuery.java
mstruk: Even better - just use a third party lib, but I didn't find any satisfactory one. Ideally I can just point to some third party library grammar, like in this case I point to the json-path/JsonPath project ...

My idea was that the number of expressions tested in JsonPathFilterQueryTest.java should serve as a source of truth or the 'language spec'.

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/JsonPathFilterQuery.java
mstruk: Implemented brackets and added short circuiting for true expressions followed by 'or'.

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/Constants.java
```diff
@@ -0,0 +1,39 @@
...
+    static final char[] LTE = {'<', '='};
+    static final char[] GTE = {'>', '='};
+    static final char[] MATCH_RE = {'=', '~'};
+    static final char[] IN = {'i', 'n'};
+    static final char[] NIN = {'n', 'i', 'n'};
+    static final char[] ANYOF = {'a', 'n', 'y', 'o', 'f'};
+    static final char[] NONEOF = {'n', 'o', 'n', 'e', 'o', 'f'};
+    //static final char[] SUBSETOF = {'s', 'u', 'b', 's', 'e', 't', 'o', 'f'};
+    //static final char[] SIZE = {'s', 'i', 'z', 'e'};
+    //static final char[] EMPTY = {'e', 'm', 'p', 't', 'y'};
```
scholzj, line 37: Should this be deleted?

scholzj (commented):
Do we have any performance numbers about the impact of the JSONPath queries? I think it should be fairly important part of this.

oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/Constants.java
mstruk: Indeed, maybe best to remove it as I don't plan to implement any of these.

mstruk:
> Do we have any performance numbers about the impact of the JSONPath queries? I think it should be fairly important part of this.

Not at the moment, but this is authentication-time check (once per session when the connection is established), and the implementation only traverses the parsed token, which we already have to parse for other reasons into JsonNode object (small maps and lists), and applying filtering according to already parsed (once at the Kafka broker startup time) JsonPath filter query / matcher. I don't see where any dramatic performance hit would come unless there is some bug in the implementation.

I can add a simple test with a bunch of loops, and print some measurement just so we have some idea of the performance. I expect it to be trivial compared to the network time required for SASL negotiation, or in the case of using introspection endpoint, to the round trip to authorization server introspection endpoint and back.

scholzj:
> Not at the moment, but this is authentication-time check (once per session when the connection is established), and the implementation only traverses the parsed token, which we already have to parse for other reasons into JsonNode object (small maps and lists), and applying filtering according to already parsed (once at the Kafka broker startup time) JsonPath filter query / matcher. I don't see where any dramatic performance hit would come from unless there is some bug in the implementation.

I can add a simple test with a bunch of loops, and print some measurement just so we have some idea of the performance. I expect it to be trivial compared to the network time required for SASL negotiation, or in the case of using introspection endpoint, to the round trip to authorization server introspection endpoint and back.

Fair enough. I guess having performance tests for the authoriser would be more important then for something what happens at authentication time only.

mstruk:
I added the test and the typical performance on my laptop of a random query is less than 20 ms to match 10000 different tokens.
The worst time is around 100 ms, and actually it could be something to look at - why a specific individual query performs 5 to 10 times worse than most of the rest. But then the improvement will not be noticed by anyone, so maybe it's not worth the time right now.

mstruk:
BTW the tests are failing due to:
```
Caused by: com.github.dockerjava.api.exception.DockerClientException: Could not pull image: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

Should I replace the image coordinates to point to quay.io where possible?

scholzj:
That depends on what images are you pulling. Not every image is on Quay.io. For the Strimzi one yes, you should be able to pull them from quay.io.

PR #84, opened 2020-12-11
fix: allow to set principal multiple times

wtrocki:
Created for testing purposes. Going to do local build


wtrocki:
FYI @mstruk

wtrocki:
Closing as there is much more work involved

PR #85, opened 2020-12-24
Refactored testsuite

mstruk:
This PR depends on #87


mstruk:
Turns out too much logging results in test failure:

    The job exceeded the maximum log length, and has been terminated.

scholzj:
Oh yeah, that is a lovely TravisCI feature :-o

.travis/build.sh
```diff
@@ -38,19 +38,27 @@ mvn spotbugs:check
 # Run testsuite with java 8 only
 if [ ${JAVA_MAJOR_VERSION} -eq 1 ] ; then
 
-  docker pull oryd/hydra:v1.0.0
+  docker pull oryd/hydra:v1.8.5
+  docker pull quay.io/keycloak/keycloak:latest
```
scholzj, line 42: Can we use some specific version instead of `:latest`?

.travis/build.sh
```diff
@@ -38,19 +38,27 @@ mvn spotbugs:check
...
+
+  clearDockerEnv
+  docker pull quay.io/strimzi/kafka:latest-kafka-2.6.0
   mvn -e -V -B clean install -f testsuite -Pkafka-2_6_0
   EXIT=$?
   exitIfError
 
   clearDockerEnv
-  docker pull strimzi/kafka:latest-kafka-2.5.1
+  docker pull quay.io/strimzi/kafka:latest-kafka-2.5.1
```
scholzj, line 61: I guess you can maybe use `0.21.1` instead of latest?

scholzj (approved):
I don't like the `latest` container images on so many places. Would prefer if it used some fixed version. But LGTM otherwise assuming the test pass of course.

.travis/build.sh
mstruk: Right, should make fore a more deterministic test env. Will change this.

.travis/build.sh
mstruk: Makes sense.

mstruk:
I'm looking into the issue with the test failing ...

mstruk:
> I'm looking into the issue with the test failing ...

Nevermind, local issue that I could only solve by doing `docker system prune`.

tombentley (approved):

PR #87, opened 2021-01-07
Custom claim check

mstruk:
This feature adds `oauth.custom.claim.check` configuration option which accepts a JSONPath filter query used to place additional constraints on the JWT tokens that are used to authenticate to your Kafka brokers.

The checking can also be performed against the introspection endpoint response.

For more about the syntax see: [JsonPathFilterQuery.java](https://github.com/mstruk/strimzi-kafka-oauth/blob/custom-claim-check-jayway/oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/JsonPathFilterQuery.java)


.travis/build.sh
```diff
@@ -55,11 +55,11 @@ if [ ${JAVA_MAJOR_VERSION} -eq 1 ] ; then
...
-  clearDockerEnv
-  docker pull strimzi/kafka:latest-kafka-2.4.1
-  mvn -e -V -B test -f testsuite -Pkafka-2_4_1
-  EXIT=$?
-  exitIfError
+#  clearDockerEnv
+#  docker pull strimzi/kafka:latest-kafka-2.4.1
+#  mvn -e -V -B test -f testsuite -Pkafka-2_4_1
+#  EXIT=$?
+#  exitIfError
```
scholzj, line 62: Should this be deleted? If not maybe you could add some explanatory comment.

README.md
```diff
@@ -218,6 +221,12 @@ If the SASL_OAUTHBEARER listener is also used for interbroker communication, the
 
     listener.name.client.oauthbearer.sasl.login.callback.handler.class=io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler 
 
+#### Enabling the custom principal builder
+
+OAuth authentication also requires a custom principal builder to be installed on the broker:
+  
+    principal.builder.class=io.strimzi.kafka.oauth.server.OAuthKafkaPrincipalBuilder
```
scholzj, line 228: How does this work for other listener using different authentication methods then OAuth?

testsuite/refresh-token-jwt-keycloak-test/src/test/resources/simplelogger.properties
```diff
@@ -0,0 +1 @@
+org.slf4j.simpleLogger.log.io.strimzi=TRACE
```
scholzj, line 1: Is this intentional? Ot some debugging left-over?

scholzj (approved):
Left some nits. But LGTM otherwise. It should be reviewed by @tombentley as well before merging.

.travis/build.sh
mstruk: I added this because now the tests run much faster which gives us room to reenable testing compatibility with older versions of Kafka for anyone who might be using this without Strimzi operator. But I can remove this from Travis CI and only run these tests locally. Is the reasoning that since we don't support these versions in the operator we also don't want to support them here? For me, I simply want to know when I do some change that breaks backwards compatibility so I can consider it, and if nothing else, make a note of it in RELEASE_NOTES.md.

README.md
mstruk: That's a good question. Broker to broker via mTLS still works. SASL_PLAIN works under the assumption that it's OAuth over PLAIN, I haven't tried with SASL_SCRAM but it's possible that it wouldn't work.
But the idea is to have a single store of users, isn't it? Do we really want to support one set of users on one listener, and a completely different unrelated set of users on another listener, and they both are then subject to a single global authorizer?

.travis/build.sh
scholzj: The reasoning is _why do we have a code in there which is commented out_. 

So ... if this is something you want to keep to uncomment and use when needed its fine. Just add above it a comment line explaining why is there and what it is useful for. This way without any explanation is not clear what it is and why it is there and why is it commented out.

testsuite/refresh-token-jwt-keycloak-test/src/test/resources/simplelogger.properties
mstruk: You mean it increases disk consumption and slows the build? I can remove it.

README.md
scholzj: > Do we really want to support one set of users on one listener, and a completely different unrelated set of users on another listener, and they both are then subject to a single global authorizer?

I think we do want to support it. there might be different groups of users - for example cluster operators with one auth mechanism on one interface, internal applications external applications etc. So there are several concerns:
* The authentication used internally for replication (today mTLS, but might be SASL in the future since there are many request to be able to avoid using TLS etc.)
* The authentication on different listeners.

The first one is a must have - that has to work. The other one is IMHO very important. But if it doesn't work we might be able to live with it but needs to be properly documented. There is also backwards compatibility to keep in mind. It is not clear to me if this is needed only for OAuth over PLAIN which is a new feature and users do not have to enable it. Or whether this affects also the OAUTHBEARER users who already might be using it in combination with other authentication methods. You could possibly also investigate if this can be configured on per-listener basis or not => as that might be a solution of course since it would not impact other listeners.

testsuite/refresh-token-jwt-keycloak-test/src/test/resources/simplelogger.properties
scholzj: No, I meant what I wrote there. I just wondered if it is intentional or some unintentional left-over. It seems a bit unusual to have the TRACE level there by default, that is all. It seems to be tests only - so if it was intentional and you think having it as TRACE by default is what we want just ignore this comment and keep it.

README.md
mstruk: The way OAuthKafkaPrincipalBuilder class is written it delegates to the DefaultKafkaPrincipalBuilder class if none of the OAuth based credentials are found. In principle it should work for any different co-existing SASL mechanisms.
It's just that there is no test for it, so it might as well not work.

testsuite/refresh-token-jwt-keycloak-test/src/test/resources/simplelogger.properties
mstruk: I see. Well the next PR that comes after this one is 'testsuite refactoring' and this module is removed. So we can as well leave this for now. But in principle if build speed and disk consumption weren't a consideration I would prefer to have logging set to TRACE all the time. Every additional iteration due to lack of info to understand why the test has failed is additional time spent, and TRACE saves that time.

oauth-common/src/test/java/io/strimzi/kafka/oauth/jsonpath/CustomCheckTest.java
```diff
@@ -0,0 +1,222 @@
...
+    /*
+     * These queries return a different result in Jayway JsonPath implementation
+     *
+     * In the case of 'anyof' and 'noneof' Jayway seems to silently fail the evaluation when the referenced
+     * attribute is missing, which results in EXPRESSION and !(EXPRESSION) to both effectively evaluate to false.
+     */
+    final String[] buggyJaywayQueries = {
+        "!(@.missing anyof [null, 'username'])", "true",
+        "!(@.missing noneof [null, 'username'])", "true"
+    };
```
tombentley, line 123: Are there issues opened against the jayway impl for these?

README.md
scholzj: I think you should test it then. Because we need to know how it does or does not work.

oauth-common/src/test/java/io/strimzi/kafka/oauth/jsonpath/CustomCheckTest.java
```diff
@@ -0,0 +1,222 @@
...
+
+            boolean result = q.matches(json);
+            System.out.println("Test: " + query + " " + (result ? "MATCH" : "NO MATCH"));
+
+            Assert.assertEquals("Expected " + expected + " when testing: " + query, expected, result);
+        }
+    }
+
+    @Test
+    public void testPerformance() throws Exception {
```
tombentley, line 177: This looks like more of a benchmark than a test: It doesn't look like it can fail. _If_ we need to keep it then perhaps we should rewrite it to work with JMH.

mstruk: Correct, it is a benchmark. It's really just to have a general idea, and I added it in order to compare with my implementation. Using JMH should rather belong to the https://github.com/json-path/JsonPath project, I don't think it makes sense to maintain it here.
I would prefer to either add @Ignore to this test or to completely remove it.

oauth-common/src/test/java/io/strimzi/kafka/oauth/jsonpath/CustomCheckTest.java
mstruk: Not yet, on my TODOs ...

mstruk: https://github.com/json-path/JsonPath/issues/662

README.md
mstruk: I added the test for this to `refactored-testsuite` branch (#85). Multiple SASL mechanisms on multiple listeners appear to work without any issues.

.travis/build.sh
mstruk: Removed it.

oauth-common/src/test/java/io/strimzi/kafka/oauth/jsonpath/CustomCheckTest.java
mstruk: Added @Ignore.

mstruk:
Is there something else to be addressed here?

README.md
mstruk: The tests are passing for #85.

scholzj:
@tombentley Were your comments addressed? Are you ok with this now?

oauth-common/src/test/java/io/strimzi/kafka/oauth/jsonpath/CustomCheckTest.java
```diff
@@ -0,0 +1,226 @@
...
+        "@.attr > null", "false",
+        "1 < @.attr", "false"
+    };
+
+    @Test
+    public void testJsonPathFilterQuery() throws Exception {
+
+        String[] errQueries = {
+            "@.attr == and !('admin' in @.roles)",
+            "@.exp < 'cant compare makes it false"
```
tombentley, line 143: Is this testing that comparing an int and a string is not allowed? Because it looks to me like the string doesn't have a closing quote.

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipal.java
```diff
@@ -45,7 +45,7 @@ public static OAuthKafkaPrincipal takeFromThreadContext() {
 
     public static void setToThreadContext(OAuthKafkaPrincipal principal) {
         if (currentTL.get() != null) {
-            throw new RuntimeException("Internal error - current principal already set");
+            throw new RuntimeException("Internal error - current principal already set. Make sure you have 'principal.builder.class=io.strimzi.kafka.oauth.server.OAuthKafkaPrincipalBuilder' in your server.properties");
```
tombentley, line 48: The config file is not _necessarily_ called `server.properties`, so maybe call it "...in your broker configuration file".

tombentley (approved):
A couple of nits, but otherwise LGTM.

oauth-common/src/test/java/io/strimzi/kafka/oauth/jsonpath/CustomCheckTest.java
mstruk: It's actually a test of unclosed quote which should generate a parsing error. I was thinking of adding some comments about the intention of every specific test, but didn't want to intersperse the lines with comments. Maybe I should add them ...

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipal.java
mstruk: We can do that.

PR #95, opened 2021-03-05
Fix for intermittent authentication failures with OAuth over PLAIN

mstruk:
OAuth over PLAIN implementation had to be slightly reworked.

This fix introduces a breaking change with 0.7.0 in how to authenticate with an access token over SASL_PLAIN by using `$accessToken` as a username.

### DIFF
--- a/README.md
+++ b/README.md
@@ -892,9 +892,9 @@ There is no OAuth specific configuration that would be required on the client wh
 The Kafka Broker has to have the SASL_PLAIN mechanism enabled and properly configured with `JaasServerOauthOverPlainValidatorCallbackHandler` validation callback handler. 
 
 Then, the standard SASL_PLAIN configuration is used on the client with the following two options:
-- the client can authenticate using the service account client ID and secret. Setting the `username` to the value of client ID, and setting the `password` to the value of client secret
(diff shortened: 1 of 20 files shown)


scholzj (approved):

README.md
```diff
@@ -892,9 +892,9 @@ There is no OAuth specific configuration that would be required on the client wh
 The Kafka Broker has to have the SASL_PLAIN mechanism enabled and properly configured with `JaasServerOauthOverPlainValidatorCallbackHandler` validation callback handler. 
 
 Then, the standard SASL_PLAIN configuration is used on the client with the following two options:
-- the client can authenticate using the service account client ID and secret. Setting the `username` to the value of client ID, and setting the `password` to the value of client secret
+- the client can authenticate using the service account client ID and secret. Setting the `username` to the value of client ID, and setting the `password` to the value of client secret. The Kafka broker will use the client credentials to obtain the access token from the authorization server.
```
michalxo, line 895: `The Kafka broker will use the client credentials` -> these client credentials?

mstruk: You mean like this?

```suggestion
- the client can authenticate using the service account client ID and secret. Setting the `username` to the value of client ID, and setting the `password` to the value of client secret. The Kafka broker will use these client credentials to obtain the access token from the authorization server. 
```

michalxo: yup. imo.,

PR #106, opened 2021-04-09
Enable ppc64le for strimzi-kafka-oauth

### LINKED ISSUE #94: Enable strimzi-kafka-oauth on ppc64le arch
Hey!

Thanks for the awesome readme/tutorials on strimzi-kafka-oauth. I could build strimzi-kafka-oauth from source & run the tests outlined, I was wondering what it would take to have the strimzi-kafka-oauth running on ppc64le. FWIW we have ppc64le instances in Travis.

krishvoor:
Enables ppc64le on strimzi-kafka-oauth CI/CD, fixes #94

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: xenial
+dist: bionic
 sudo: required
 language: java
 jobs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,13 @@ jobs:
       apt:
         packages:
         - maven
+  - os: linux
+    arch: ppc64le
+    jdk: openjdk11
+    addons:
+      apt:
+        packages:
+        - maven
 jdk:
 - openjdk8
 - openjdk11


krishvoor:
FWIW,
We have only enabled JDK11 bases tests, as dependent docker-images for JDK8 are being worked upon.

scholzj (approved):

sknot-rh (approved):

mstruk:
Looks good, just keep in mind that the testsuite only runs when Java 8 is used for the build.
Since you are using Java 11, the testsuite is skipped in your case - only unit tests are executed, but not a separate 'teststuite' submodule.

If you want to run the 'testsuite' as part of your build change [this line](https://github.com/strimzi/strimzi-kafka-oauth/blob/main/.travis/build.sh#L39) to:

    if [ ${JAVA_MAJOR_VERSION} -ne 1 ] ; then

mstruk (approved):

scholzj:
@mstruk The problem is that the test suite is using container images which would need to be available for ppc64le ... and not all of them are. So that would need to be addressed first before enabling the tests.

scholzj:
@krishvoor Thanks for the PR.

PR #102, opened 2021-03-29
Fix examples and improve docs

mstruk:

PR #107, opened 2021-04-13
Add support in 'OAuth over PLAIN' for mode where token is passed without '$accessToken:' prefix

mstruk:
Add a non-breaking addition whereby not setting the `oauth.token.endpoint.uri` on the listener, the `OAuth over PLAIN` works in 'access-token-only' mode (or 'no-client-credentials' mode), where `username` and `password` parameters of PLAIN authentication are always treated as account id + access token, never as Client ID + secret. In this mode the value of `password` parameter should never be prefixed by '$accessToken:' to signify that access token is passed. Rather, in this mode it is always assumed that the access token is passed as-is.

If `oauth.token.endpoint.uri` is configured, then the current behaviour stays the way it currently is - the default is to interpret the `username` and `password` parameters as Client ID + secret, unless the '$accessToken:' prefix is detected in which case the parameters are interpreted as account id + access token.

This PR is a continuation of #103.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,12 +24,13 @@ Strimzi Kafka OAuth modules provide support for OAuth2 as authentication mechani
     - [Configuring the listeners](#configuring-the-listeners)
     - [Configuring the JAAS login module](#configuring-the-jaas-login-module)
     - [Enabling the custom callbacks](#enabling-the-custom-callbacks)
-    - [Enabling the custom principal builder](#enabling-the-custom-principal_builder)
+    - [Enabling the custom principal builder](#enabling-the-custom-principal-builder)
     - [Configuring the OAuth2](#configuring-the-oauth2)
       - [Configuring the token validation](#configuring-the-token-validation)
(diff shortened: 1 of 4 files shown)


scholzj (approved):

elakito:
@mstruk Looks good. Thanks.

tombentley (approved):

scholzj:
@mstruk I guess we can merge this, right? Should we do 0.8.0 release afterwards? Or include it in 0.7.2?

mstruk:
It would make sense to do 0.7.2 indeed as OAuth over PLAIN functionality would then neatly be completed in 0.7.x branch.
We can merge this, but then It would be good to also merge #102 before releasing 0.7.2.

scholzj:
@mstruk Ok. Please have a look at #102 then ... so that we can release it and merge the PR in operators repo. Thanks.

mstruk:
Do we add Milestone 0.7.2 so we can properly tag it as such? I pushed all the updated for #102, it's running tests now ...

PR #111, opened 2021-05-31
Replace keycloak helper library with nimbus-jose-jwt

mstruk:
This completely removes dependency on `keycloak-*` jars and on BouncyCastle library. ECDSA support is now enabled by default using Java JCE rather than BouncyCastle. This also fixes edge cases where JWT tokens contain certain semi-standard claims with value types not compatible with `keycloak-*` libraries' expectations.

As a result the options `oauth.crypto.provider.bouncycastle` and `oauth.crypto.provider.bouncycastle.position` have been deprecated, and are ignored if specified.


scholzj (approved):

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
```diff
@@ -254,16 +228,34 @@ private PublicKey getKeyUnlessStale(String id) {
...
+
             newCache = Collections.unmodifiableMap(newCache);
             if (!cache.equals(newCache)) {
                 log.info("JWKS keys change detected. Keys updated.");
                 oldCache = cache;
                 cache = newCache;
             }
             lastFetchTime = System.currentTimeMillis();
-        } catch (Exception ex) {
+        } catch (Throwable ex) {
```
tombentley, line 258: What non-Exception can be thrown here now?

mstruk: There are no additional exceptions / errors that could be thrown here. I just noticed, that non-exceptions can make it through uncaught, potentially avoiding the log or making the log entry without the information about the SSO endpoint url that was affected.

tombentley (approved):

PR #113, opened 2021-06-07
Add support for PEM type truststores to the OAuth client

scholzj:
Right now, the client requires that the truststore for the HTTPS communication between the OAuth client (both in broker and in Kafka client) and the OAuth server is passed as JKS or PKCS12 formats. But in many cases, you have the certificate only in the PEM format. So you need to convert it first for example using Java keytool or OpenSSL. 

This PR adds support to using PEM files directly, without any need for some pre-processing. That should make it much easier to use in environments such as Kubernetes where no special startup scripts should be needed anymore. It follows the way Apache Kafka supports them - they can be passed both as path / location or directly in one of the configuration options without using any files.

The Hydra tests which already use TLS have now additional tests to test this feature as well. (Some additional changes tot he tests were needed to ensure the isolation between the different tests)


oauth-common/src/main/java/io/strimzi/kafka/oauth/common/SSLUtil.java
```diff
@@ -27,11 +29,28 @@ public static SSLSocketFactory createSSLFactory(String truststore, String passwo
...
-            store.load(is, password.toCharArray());
-        } catch (Exception e) {
-            throw new RuntimeException("Failed to load truststore: " + truststore, e);
+
+        if ("PEM".equals(type)) {
+            try (FileInputStream is = new FileInputStream(truststore)) {
+                store = KeyStore.getInstance("PKCS12");
+                store.load(null, null);
+
+                while (is.available() > 0) {
```
mstruk, line 38: FileInputStream.available() is not meant to be used this way - it can return 0 at any time, depending on the filesystem ...
(https://stackoverflow.com/questions/2788869/may-the-fileinputstream-available-foolish-me)

If using the code example from [here](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/security/cert/CertificateFactory.html) then it's really important (as hinted in the last sentence before the example) to wrap the FileInputStream with a BufferedInputStream.

```
        if ("PEM".equals(type)) {
            try (BufferedInputStream is = new BufferedInputStream(new FileInputStream(truststore)) {
```

mstruk (changes requested):
See [comment](https://github.com/strimzi/strimzi-kafka-oauth/pull/113/files#r646725786)

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/SSLUtil.java
scholzj: I didn't used that example, but it is a good point. I updated the code.

mstruk (approved):

scholzj:
@mstruk Sorry, I reworked this a bit and added also an option to pass the certificate data directly in the options. Can you please have another look?

RELEASE_NOTES.md
```diff
@@ -4,6 +4,11 @@ Release Notes
 0.8.0
 -----
 
+### Support for PEM certificates
+
+PEM certificates can be now used directly without being converted to Java Keystore or PKCS12 formats.
```
mstruk, line 9:
```suggestion
PEM certificates can now be used directly without being converted to Java Keystore or PKCS12 formats.
```

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/ConfigProperties.java
```diff
@@ -30,6 +30,13 @@ public static void resolveAndExportToSystemProperties(Properties defaults) {
         }
     }
 
+    public static void clearSystemProperties(Properties defaults) {
+        Properties p = new ConfigProperties(defaults).resolveTo(new Properties());
+        for (Object key: p.keySet()) {
+            System.clearProperty(key.toString());
+        }
+    }
+
```
mstruk, line 39: I would prefer the above method to not be part of oauth-common but just be included in the hydra-test as there it's the only place where it is used.

testsuite/kafka.log
```diff
@@ -0,0 +1,1162 @@
+#
```
mstruk, line 1: This file shouldn't be in the git. It simplifies debugging that the output is dumped into root dir.
I'll add it to .gitignore file to prevent it from being automatically added for commit by IDEs.

mstruk (commented):
Just a few nits. You can fix it now or we can merge, and I can fix some of these things later.

scholzj:
@mstruk The comments should be fixed now.

mstruk (approved):
Nice, thanks.

PR #125, opened 2021-09-13
Fix for occasional occurrence of wrong principal type with OAuth over PLAIN + tests

mstruk:
The same SaslServer can call the principal builder multiple times. Previously the assumption was that that would never happen.


scholzj (approved):

tombentley (commented):
Is the FloodTest really the only way we can test this? Given that that seems to rely on thread scheduling to actually detect a problem it seems like it's not a regression test that can be relied on.

mstruk:
This test allowed me to reproduce an issue so I was able to start fixing it, and I had no problem reproducing the issue with this test. While it may be runtime specific, it has value in that if there is still some issue with false assumptions it should sooner or later be identified by seeing this test fail in some runtime environment.

The alternative would be a lot of mocking of Kafka context which would be complex and quite possibly buggy in terms of properly simulating the underlying Kafka threading behavior.

I prefer end-to-end integration test to a mocked up unit test. It has more chance to catch any change in behavior introduced in Kafka compared to a test that tries to meticulously simulate current Kafka behaviour.

Or do you maybe have some third approach in mind?

tombentley:
I'm not saying that the FloodTest doesn't have value. I am saying that an approach where we only rely on such a test is going to be flaky. I don't know this code particularly well, but I just wrote this:

```java
class OAuthKafkaPrincipalBuilderTest {

    @Test
    public void testPreviousStoredPrincipalIsReused() {
        Services.configure(Collections.emptyMap());
        OAuthKafkaPrincipalBuilder builder = new OAuthKafkaPrincipalBuilder();
        PlainSaslServer saslServer = Mockito.mock(PlainSaslServer.class);
        Mockito.when(saslServer.getAuthorizationID()).thenReturn("tom");
        SaslAuthenticationContext context = Mockito.mock(SaslAuthenticationContext.class);
        Mockito.when(context.server()).thenReturn(saslServer);

        OAuthKafkaPrincipal storedPrincipal = new OAuthKafkaPrincipal(KafkaPrincipal.USER_TYPE, "tom");
        Services.getInstance().getCredentials().storeCredentials("tom", storedPrincipal);

        assertTrue(storedPrincipal == builder.build(context));
    }
}
```
which would be enough of a regression test, since AFAICS the fix you've implemented is not really about the threading behaviour as such. I don't think that test is especially hard, or bug-prone.

tombentley (approved):

PR #132, opened 2021-11-12
Add connect and read timeouts for communication with authorization server

mstruk:
In some circumstances the network `glitches` can result in indefinite blocking of JWKS keys refresh background job. This commit adds setting the connect and read timeouts with default value of 60 seconds for both. Use `oauth.connect.timeout.seconds` and `oauth.read.timeout.seconds` system properties, env vars or as part of listener or client jaas config to set non-default values in seconds.


oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
```diff
@@ -23,6 +23,8 @@
     public static final String OAUTH_SSL_SECURE_RANDOM_IMPLEMENTATION = "oauth.ssl.secure.random.implementation";
     public static final String OAUTH_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM = "oauth.ssl.endpoint.identification.algorithm";
     public static final String OAUTH_ACCESS_TOKEN_IS_JWT = "oauth.access.token.is.jwt";
+    public static final String OAUTH_CONNECT_TIMEOUT = "oauth.connect.timeout";
+    public static final String OAUTH_READ_TIMEOUT = "oauth.read.timeout";
```
scholzj, line 27: Would it make sense to add the unit into the option name? E.g. `oauth.connect.timeout.seconds` and `oauth.read.timeout.seconds`? It makes it more readable and both Kafka and Strimzi seem to use this in many places.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
```diff
@@ -142,6 +163,32 @@ public static void delete(URI uri, SSLSocketFactory socketFactory, HostnameVerif
...
+    private static void configureTimeouts(HttpURLConnection con, int connectTimeout, int readTimeout) {
+        if (connectTimeout <= 0) {
+            throw new IllegalArgumentException("connectTimeout <= 0");
+        }
+        con.setConnectTimeout(connectTimeout * 1000);
+
+        if (readTimeout <= 0) {
+            throw new IllegalArgumentException("readTimeout <= 0");
+        }
+        con.setReadTimeout(readTimeout * 1000);
```
scholzj, line 189: Should the options be in milliseconds if milliseconds are used here?

testsuite/keycloak-errors-tests/src/test/java/io/strimzi/testsuite/oauth/auth/Common.java
```diff
@@ -144,4 +150,36 @@ static String loginWithUsernamePassword(URI tokenEndpointUri, String username, S
         }
         return token.asText();
     }
+
+    public static List<String> getKafkaLogsForError(String errString) {
```
scholzj, line 154: Maybe some Javadoc to explain what it does?

scholzj (commented):
Some higher-level questions. The code looks good to me.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
mstruk: Actually it makes a lot of sense since there are already `oauth.jwks.refresh.seconds`, `oauth.jwks.expiry.seconds`, and `oauth.jwks.refresh.min.pause.seconds`.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
mstruk: Not sure anyone would want to be that precise about it. For one, having the minimum timeout of 1s prevents clients from flooding the authorization server with requests (for example by setting read timeout to 1ms). Primarily the idea is to be able to set the timeout and retry a request if nothing happens for 30 seconds for example. It makes no difference if it is 30s or 30.149s. I don't see why someone would want to set the connect timeout or read timeout to less than 1 second for example. But maybe there is a use case, and I just don't see it.

testsuite/keycloak-errors-tests/src/test/java/io/strimzi/testsuite/oauth/auth/Common.java
mstruk: Sure.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
mstruk: Also, maybe 60 seconds is not the best default for connect timeout and read timeout. Maybe 30 seconds makes more sense? Or maybe read timeout is better to be different than the connect timeout?

Primarily the idea is to not have infinite timeout as that may result in a `hijacked` request thread. It's when network layer has issues, or the server is not responding. The question is when did we wait enough, and when is it time to give up on the request. Exponential backoff is implemented separately from that, but can't work properly if its thread is `hijacked`.

scholzj: I think it is common to us milliseconds as the unit even through you usually set it only to seconds. But if we already have other options in seconds here I guess we should stuck with it.

scholzj (approved):

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/HttpUtil.java
scholzj: > Also, maybe 60 seconds is not the best default for connect timeout and read timeout. Maybe 30 seconds makes more sense? Or maybe read timeout is better to be different than the connect timeout?

> Primarily the idea is to not have infinite timeout as that may result in a hijacked request thread. It's when network layer has issues, or the server is not responding. The question is when did we wait enough, and when is it time to give up on the request. Exponential backoff is implemented separately from that, but can't work properly if its thread is hijacked.

I think these are good questions. But not sure I have the answers I'm afraid.

PR #133, opened 2021-11-22
Add groups extraction and expose them via OAuthKafkaPrincipal

mstruk:


.checkstyle/checkstyle.xml
```diff
@@ -90,7 +90,7 @@
         <module name="MethodLength"/>
         <module name="ParameterNumber">
             <!-- default is 8 -->
-            <property name="max" value="13"/>
+            <property name="max" value="15"/>
```
scholzj, line 93: Isn't it better to suppress the warning with annotation on the class where you really need it instead of bumping the limit for all classes here?

mstruk: I can also do it like that yes.

jonathan-albrecht-ibm:
I work on s390x kafka related stuff and I happened to notice the failed travis s390x build for this PR. I had a look into it and just wanted to drop a quick note about it in case its helpful.

This is the failure from travis s390x:
```
[ERROR] testHttpTimeouts(io.strimzi.kafka.oauth.common.HttpUtilTimeoutTest)  Time elapsed: 12.038 s  <<< FAILURE!
java.lang.AssertionError: Unexpected diff: 2000
	at io.strimzi.kafka.oauth.common.HttpUtilTimeoutTest.testHttpTimeouts(HttpUtilTimeoutTest.java:74)
```
Note the `Unexpected diff` value is 2000. This is the failing part of the test:
```
            timeout = 2;
            try {
                start = System.currentTimeMillis();
                HttpUtil.get(URI.create("http://172.0.0.13:8079"), null, null, null, String.class, timeout, timeout);

                Assert.fail("Should fail with SocketTimeoutException");
            } catch (SocketTimeoutException e) {
                long diff = System.currentTimeMillis() - start;
                Assert.assertTrue("Unexpected error: " + e, e.toString().contains("connect timed out"));
                Assert.assertTrue("Unexpected diff: " + diff, diff > timeout * 1000 && diff < timeout * 1000 + 1000);
            }
```
The last assertTrue is failing because `diff == 2000` and `timeout * 1000 == 2000` so `2000 > 2000` is false.

If I have that right, I think the test shouldn't be failing in this case and I don't think its s390x specific. Maybe the first part of the diff check should be something like: `diff >= timeout * 1000` or maybe `diff > timeout * 1000 - 10` to give a little extra room.

mstruk:
@jonathan-albrecht-ibm I noticed this one as well. It's related to #132. Based on the error the diff is exactly 2000 ms which does not satisfy `diff > timeout * 1000`. It's a buggy condition in the test.

mstruk:
This PR depends on #134.

scholzj (approved):

mstruk:
@tombentley Just a friendly reminder to find some time to review this if you can. Thanks.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/BearerTokenWithPayload.java
```diff
@@ -25,6 +27,10 @@
 
     void setPayload(Object payload);
 
+    List<String> getGroups();
+
+    ObjectNode getJSON();
```
tombentley, line 32: Can we add Javadoc to these?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/JSONUtil.java
```diff
@@ -87,18 +89,23 @@ public static String getClaimFromJWT(JsonNode node, String... path) {
...
+
+    public static List<String> asListOfString(JsonNode arrayOrString, String delimiter) {
 
         ArrayList<String> result = new ArrayList<>();
 
-        if (arrayNode.isTextual()) {
-            result.addAll(Arrays.asList(arrayNode.asText().split(" ")));
+        if (arrayOrString.isTextual()) {
+            result.addAll(Arrays.asList(arrayOrString.asText().split(Pattern.quote(delimiter)))
+                    .stream().map(String::trim).collect(Collectors.toList()));
```
tombentley, line 102: Do we want to filter out empty strings (which would happen if you had two delimiters next to each other)?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/BearerTokenWithPayload.java
```diff
@@ -25,6 +27,10 @@
 
     void setPayload(Object payload);
 
+    List<String> getGroups();
```
tombentley, line 30: Isn't it logically a `Set`?

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ValidatorKey.java
```diff
@@ -75,7 +84,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return Objects.hash(validIssuerUri, audience, customClaimCheck, usernameClaim, fallbackUsernameClaim, fallbackUsernamePrefix, sslTruststore, sslStorePassword, sslStoreType, sslRandom, hasHostnameVerifier, connectTimeout, readTimeout);
+        return Objects.hash(validIssuerUri, audience, customClaimCheck, usernameClaim, fallbackUsernameClaim, fallbackUsernamePrefix, groupQuery, groupDelimiter, sslTruststore, sslStorePassword, sslStoreType, sslRandom, hasHostnameVerifier, connectTimeout, readTimeout);
```
tombentley, line 87: Putting the arguments on separate lines would make it easier to read.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ValidatorKey.java
```diff
@@ -108,7 +119,7 @@ public JwtValidatorKey(String validIssuerUri,
                                int connectTimeout,
                                int readTimeout) {
 
-            super(validIssuerUri, audience, customClaimCheck, usernameClaim, fallbackUsernameClaim, fallbackUsernamePrefix, sslTruststore, sslStorePassword, sslStoreType, sslRandom, hasHostnameVerifier, connectTimeout, readTimeout);
+            super(validIssuerUri, audience, customClaimCheck, usernameClaim, fallbackUsernameClaim, fallbackUsernamePrefix, groupQuery, groupDelimiter, sslTruststore, sslStorePassword, sslStoreType, sslRandom, hasHostnameVerifier, connectTimeout, readTimeout);
```
tombentley, line 122: Again, painful long line.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ValidatorKey.java
```diff
@@ -164,7 +177,7 @@ public IntrospectionValidatorKey(String validIssuerUri,
                                   int connectTimeout,
                                   int readTimeout) {
 
-            super(validIssuerUri, audience, customClaimCheck, usernameClaim, fallbackUsernameClaim, fallbackUsernamePrefix, sslTruststore, sslStorePassword, sslStoreType, sslRandom, hasHostnameVerifier, connectTimeout, readTimeout);
+            super(validIssuerUri, audience, customClaimCheck, usernameClaim, fallbackUsernameClaim, fallbackUsernamePrefix, groupQuery, groupDelimiter, sslTruststore, sslStorePassword, sslStoreType, sslRandom, hasHostnameVerifier, connectTimeout, readTimeout);
```
tombentley, line 180: And again.

tombentley (commented):
Sorry @mstruk, this one dropped off my radar. Thanks for the reminder.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/BearerTokenWithPayload.java
mstruk: I am unsure about this one as well.

That depends on the semantics. Mostly people will want it to have a Set semantics, but the authorization server may maintain predictable order, and in principle it may allow configuring the order (I don't know how common that is). In such a case, since JSON array or a string is always used to communicate the info, the ordering is maintained all the way to the BearerTokenWithPayload object, and could possibly be used when making authorization decision in a custom authorizer. While I have no useful use-case in mind, maybe someone else has - at their peril. Thus, I see this as an additional degree of freedom, that shouldn't really hurt anyone.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/BearerTokenWithPayload.java
mstruk: Added some JavaDoc.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/JSONUtil.java
mstruk: This method is a general purpose method, so we probably don't want to remove empty strings there. The groups extraction usage makes sure to ignore empty strings.  
See:
- https://github.com/strimzi/strimzi-kafka-oauth/pull/133/files#diff-1af00685f38f1cd3d7e842281c5388d45ca757226cf5cf87bf81c41c597f1a0eR396
- https://github.com/strimzi/strimzi-kafka-oauth/pull/133/files#diff-c6b310d93d0ab3ac6eae52a71596528d2a1fc2f31b6def041735574ca443305bR282

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/BearerTokenWithPayload.java
mstruk: Since most people I talked to expect a Set here I'm changing to Set.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/JSONUtil.java
tombentley: OK, but if it's really supposed to be general purpose you should define the contract in some javadoc. The possibility of delims next to each other is easily overlooked (in the common case where you're expecting a list of non-empty tokens).

README.md
```diff
@@ -454,6 +454,37 @@ For example:
 
 See [JsonPathFilterQuery JavaDoc](oauth-common/src/main/java/io/strimzi/kafka/oauth/jsonpath/JsonPathFilterQuery.java) for more information about the syntax.
 
+###### Group extraction
+
+When using custom authorization (by installing a custom authorizer) you may want to take user's group membership into account when making the authorization decisions.
+One way is to obtain and inspect a parsed JWT token from `io.strimzi.kafka.oauth.server.OAuthKafkaPrincipal` object available through `AuthorizableRequestContext` passed to your `authorize()` method.
+Another way is to configure group extraction at authentication time, and get groups as a set of principals from `OAuthKafkaPrincipal` object.
+There are two configuration parameters for configuring group extraction:
```
tombentley, line 462: Should we provide some guidance about when one is preferable to the other?

tombentley (approved):
A couple of minor points, but LGTM.

README.md
mstruk: This just depends on what the custom authorizer requirements are. It may be that the developer just needs a list of groups to make the authorization decision. Or it may need some other claims that are only available in the JWT token. I don't really have an idea what more to add.

mstruk: I did rework the text a little bit.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/JSONUtil.java
mstruk: Added JavaDoc to JSONUtil.java

tombentley (approved):

PR #136, opened 2022-02-12
Added zk dependency to kafka and restart policy to compose files

iMajna:
While deploying Kafka with Hydra there are times where Kafka shuts down because Zookeeper just wasn't ready to accept connections. 
To tweak it just a little bit I added small dependency to Kafka + restart policy just in case it dies for some unknown reason to tackle "Did you try to turn it off and on again".


scholzj (commented):
Thanks for the PR. Could you please fix the DCO sign-off? the instructions should be under the _details_ link next to the DCO status. Thanks.

iMajna:
Ah, my bad, apologise!
This should do it :)

PR #141, opened 2022-03-18
OAuth metrics

mstruk:
This PR adds JMX MBeans for tracking OAuth related metrics on Kafka brokers and clients, including KeycloakRBACAuthorizer metrics.


oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ValidatorKey.java
```diff
@@ -58,6 +64,24 @@
         this.hasHostnameVerifier = hasHostnameVerifier;
         this.connectTimeout = connectTimeout;
         this.readTimeout = readTimeout;
+        this.enableMetrics = enableMetrics;
+
+        this.configIdHash = IOUtil.hashForObjects(validIssuerUri,
```
scholzj, line 69: I'm a bit puzzled what is this used for and how it relates to the metrics. Can you elaborate on it?

scholzj (commented):
I left few comments here and there. But TBH given the size of it and my knowledge of the code base, I'm not sure I can do much more detailed review. So I guess as long as it works ...

Not sure if @tombentley has to to look at it - it might be useful, but not sure if he adds more value than me.

.travis/build.sh
```diff
@@ -69,11 +69,12 @@ if [ ${JAVA_MAJOR_VERSION} -eq 1 ] ; then
...
-  docker pull quay.io/strimzi/kafka:0.23.0-kafka-2.6.2
-  mvn -e -V -B clean install -f testsuite -Pkafka-2_6_2
-  EXIT=$?
-  exitIfError
+#  Remove this very old version from build due to the whole thing taking too long
+#  clearDockerEnv
+#  docker pull quay.io/strimzi/kafka:0.23.0-kafka-2.6.2
+#  mvn -e -V -B clean install -f testsuite -Pkafka-2_6_2
+#  EXIT=$?
+#  exitIfError
```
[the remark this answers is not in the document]
mstruk: Maybe best to just delete it yes.

examples/kubernetes/kafka-oauth-single-authz-metrics.yaml
```diff
@@ -0,0 +1,336 @@
...
+      tokenEndpointUri: http://keycloak:8080/auth/realms/kafka-authz/protocol/openid-connect/token
+      delegateToKafkaAcls: true
+      superUsers:
+      - User:service-account-kafka
+    logging:
+      type: inline
+      loggers:
+        log4j.logger.io.strimzi: "TRACE"
+        log4j.logger.kafka: "DEBUG"
+        log4j.logger.org.apache.kafka: "DEBUG"
```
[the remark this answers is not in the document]
mstruk: It's just an example, but if you think that it's excessive logging I can remove / set different level.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ValidatorKey.java
mstruk: When there are many listeners configured it results in multiple instances of the validator callback handler and consequently of the validators. I introduced an `oauth.config.id` config parameter to allow the user to set the name for the listener - it does not seem to be possible to get the listener name from the custom ValidateCallbackHandler. You want to segment the metrics by validators (listeners) rather than have it all merged together. And you need a discriminator (a context id) to qualify metrics. Thus if `oauth.config.id` was not set, it is calculated from other config parameters, resulting in a short hashstring so that every validator has a config id that is included as an attribute in the metric. When you map this to Prometheus JMX Exporter, you can then do PromQL queries on a per validator basis or you automatically get segmentation on a per validator basis.

mstruk:
> Not sure if @tombentley has to to look at it - it might be useful, but not sure if he adds more value than me.

In my experience @tombentley always finds something to improve the PR 😀

scholzj (approved):

mimaison:
I really don't have the background for this item but my question is:
If as a user I have defined a metrics reporter for my Kafka application (by setting `metric.reporters` on a producer for example), will it get the metrics this plugin registers? or will these metrics only be accessible via JMX?

mstruk:
The OAuth metrics are exported to JMX independently and Kafka metrics API is not used. Therefore, as things stand, the `metric.reporters` config will have no effect on them. They are enabled independently through `oauth.enable.metrics`. I guess that config is something you as a user prefer to be integrated as it makes for a more unified experience? Are there other benefits from such integration?

mimaison:
I believe frameworks like Quarkus retrieve metrics via the `metrics()` API Tom mentioned above. This API is backed by the same mechanism as `metric.reporter`, so I think there's value in properly integrating with the Kafka metrics.

mstruk:
It looks like [Metrics](https://kafka.apache.org/31/javadoc/org/apache/kafka/common/metrics/Metrics.html) object which serves as a singleton and an API entrypoint to Metrics API is held as an instance variable in `KafkaServer` instance. That instance in turn is not stored in some other registry or in some way reachable from a custom `AuthenticateCallbackHandler` (which is a plugin interface for custom handling of authentication). Similarly there is no way to get to the `KafkaServer` from the custom `Authorizer` (which we implement with `KeycloakRBACAuthorizer`).

As I see it, without access to `KafkaServer` instance there doesn't seem to be a way to plug into the common registry of registered metrics. While we can use the Metrics API, it will be just a parallel object tree in the memory, out of reach to mechanisms handled by Kafka based on standard config (handled by the `KafkaServer` instance). 

`Metrics` object is easy to get from the `KafkaProducer` and the `KafkaConsumer`, but again - not from within the LoginCallbackHandler where our code is bootstrapped. The metrics object would have to be manually retrieved by client and passed into our context. Something like:

```
KafkaProducer producer = new KafkaProducer(...);
Services.instance().initMetrics(producer.metrics());
```

Where `Services.instance()` is part of Strimzi OAuth, and the above code is executed by the developer writing their Kafka client. Without that call we would again be unable to integrate with metrics config and mechanisms provided by `KafkaProducer` / `KafkaConsumer`.

It appears that Kafka Metrics is a walled garden, not intended to be meddled with by extensions.

mstruk:
I added an integration with `metric.reporters` configuration, and migrated from custom gauge / metric classes to using `org.apache.kafka.common.metrics` package.

The current integration works around the current inability to get to existing `Metrics` object used by whatever runtime bootstraps the JAAS configuration. Also it works around the current inability to obtain a list of configured `MetricReporter`s used by the same runtime. As a consequence we parse the same configuration, and initialise and configure the same objects as the current runtime does (whatever the runtime is).

When I say runtime I mean Kafka broker, Kafka client (producer / consumer / admin), Kafka connect, Kafka streams ...
Each of these runtimes instantiates its own `Metrics` object, and its own list of `MetricReporter`s.
As a result there will be an additional instance of `Metrics` and an additional list of `MetricReporter` instances within the same application process.

There is some prototyping going on by @mimaison in Kafka code base towards improving the ability to integrate with Kafka metrics so that in the future we could obtain and use an existing `Metrics` object already configured by the current runtime. 
Additionally we might want to obtain a list of `MetricReporter`s already configured by the current runtime.

There are some considerations when deciding whether to instantiate and configure our own `Metrics` object, or use an existing one.

A `Metrics` object has a single namespace which depending on the runtime is `kafka.server`, `kafka.producer`, `kafka.consumer`, `kafka.connect`, `kafka.streams` ... That namespace becomes the first part of the JMX MBean names for the metrics as exported to JMX by `JmxReporter`.

For OAuth metrics we might want to use our own - for example `strimzi.oauth`. Current implementation does that. Currently we can only guess about the environment, but can't get an already configured `Metrics` instance or the `MetricsContext` instance.

There are several config options in Kafka that control `Metrics` configuration - `metrics.num.samples`, `metrics.recording.level`, `metrics.sample.window.ms`. We have to parse these from the config ourselves even though they were already parsed and interpreted by the current runtime. If we had access to the runtime's `MetricsContext` and `MetricConfig` objects we could avoid that step.

Also, in order to integrate with `metric.reporters` configuration we either need access to the already parsed list of `MetricReporter` instances, or we have to parse the configuration and instantiate the `MetricReporter` instances ourselves. Doing it that way results in double the number of `MetricReporter` instances in the process. The first set, instantiated by the runtime, is connected to and handles the events of the runtime's `Metrics` object, and the second set is instantiated by us, and handles the events from our `Metrics` instance. There is no double handling of the metrics events, but it would make more sense to have a single list of `MetricReporter`s handle events for both `Metrics` objects.

In light of the above the runtime could in the future provide integrators with multiple objects, and letting the integrator decide what to use.

For example:

```
interface MetricsIntegration {
    void setMetricsContext(MetricsContext ctx);
    void setMetricConfig(MetricConfig cfg);
    void setMetricReporters(List<MetricReporter> reporters);
    void setMetrics(Metrics metrics);
} 
```

mstruk:
WRT 'KIP-714 Client metrics and observability', I could find no branch or PR with a prototype implementation, but based on the specification alone there is no client side integration defined, thus if we create our own Metrics object the way current PR implementation does, it will most certainly not be picked up by client metrics push infrastructure. The only way to go would be to work with the Metrics object provided by the runtime.

In order to support the most flexibility for integrators I think the implementation of KIP-714 in Kafka Clients java library would have to define some integration mechanism for extensions that want to use their own Metrics object.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
```diff
@@ -0,0 +1,179 @@
...
+                        ConfigDef.Importance.LOW,
+                        CommonClientConfigs.METRIC_REPORTER_CLASSES_DOC);
+
+        return new AbstractConfig(configDef, configMap);
+    }
+
+    private KafkaMetricsContext createKafkaMetricsContext() {
+        HashMap<String, String> contextLabels = new HashMap<>();
+
+        // We don't know how to get clusterId here
```
mimaison, line 102: Should we mention the file/class/package that contains this code?

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private Metrics createMetrics(MetricConfig metricConfig, KafkaMetricsContext ctx, List<MetricsReporter> reporters) {
+        log.debug("Creating Metrics: \n    metricConfig: {samples: " + metricConfig.samples() + ", recordingLevel: " +
+                metricConfig.recordLevel() + ", timeWindowMs: " + metricConfig.timeWindowMs() + "}\n    kafkaMetricsContext: " +
+                ctx.contextLabels() + "\n    reporters: " + reporters);
+        return new Metrics(metricConfig, reporters, Time.SYSTEM, ctx);
+    }
+
+    private void addMetricsToSensor(Metrics metrics, Sensor sensor, MetricKey key) {
+        MetricName metricName = new MetricName("count", key.getName(), "Total equest count", key.getAttrs());
```
mimaison, line 164: `equest` -> `request`

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
```diff
@@ -0,0 +1,179 @@
...
+            }
+        } else {
+            String nodeId = config.getValue("node.id");
+            if (nodeId != null) {
+                contextLabels.put("kafka.node.id", nodeId);
+            }
+        }
+
+        if (config.getValue("client.id") != null) {
+            String clientId = config.getValue("client.id");
```
mimaison, line 126: Should we use `CommonClientConfigs.CLIENT_ID_CONFIG` instead of `client.id`?

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
```diff
@@ -0,0 +1,179 @@
...
+        for (Map.Entry<String, ?> entry: configMap.entrySet()) {
+            if (entry.getKey().startsWith(prefix)) {
+                contextLabels.put(entry.getKey().substring(prefix.length()), String.valueOf(entry.getValue()));
+            }
+        }
+    }
+
+    private MetricConfig getMetricConfig() {
+
+        int numSamples = config.getValueAsInt("metrics.num.samples", 2);
```
mimaison, line 147: We could use the `CommonClientConfigs` constants here too

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ValidatorKey.java
mimaison: Could we use `Objects.hash()` instead of all this custom code?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
```diff
@@ -474,16 +507,21 @@ private boolean denyIfTokenInvalid(BearerTokenWithPayload token) {
...
         try {
             grants = fetchAuthorizationGrants(token.value());
             if (grants == null) {
                 grants = JSONUtil.newObjectNode();
             }
-        } catch (HttpException e) {
-            if (e.getStatus() == 403) {
-                grants = JSONUtil.newObjectNode();
+        } catch (Throwable e) {
+            if (e instanceof HttpException) {
```
mimaison, line 517: If we're only interested in handling `HttpException`, let's catch `Exception` instead of `Throwable`

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
```diff
@@ -302,7 +355,9 @@ private void fetchKeys() {
                 cache = newCache;
             }
             lastFetchTime = System.currentTimeMillis();
+
         } catch (Throwable ex) {
+            addJwksHttpMetricErrorTime(ex, requestStartTime);
```
mimaison, line 360: Catching a `Throwable` and rethrowing it as `RuntimeException` is a bit weird. I'm not familiar with this code but is it really what we want to do?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
```diff
@@ -405,17 +436,20 @@ public List<AuthorizationResult> authorize(AuthorizableRequestContext requestCon
...
+                result = allowOrDenyBasedOnGrants(requestContext, actions, grants);
+            } else {
+                result = delegateIfRequested(requestContext, actions, null);
             }
-
-            return delegateIfRequested(requestContext, actions, null);
+            addAuthzMetricSuccessTime(startTime);
+            return result;
 
         } catch (Throwable t) {
```
mimaison, line 446: We should not catch and swallow `Throwable`

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/metrics/GrantsHttpMetricKeyProducer.java
```diff
@@ -0,0 +1,33 @@
...
+public class GrantsHttpMetricKeyProducer extends AbstractMetricKeyProducer {
+
+    public GrantsHttpMetricKeyProducer(String contextId, URI uri) {
+        super(contextId, uri);
+    }
+
+    @Override
+    public MetricKey successKey() {
+        Map<String, String> attrs = JmxMetrics.getMetricKeyAttrs(contextId, uri, "keycloak-authorization");
+        attrs.put("outcome", "success");
```
mimaison, line 23: We have this code so many times that I wonder if we should have a `addSuccessAttrs()` method, or even `getSuccessMetricKeyAttrs()`, WDYT?

testsuite/common/src/main/java/io/strimzi/testsuite/oauth/common/TestMetrics.java
```diff
@@ -0,0 +1,154 @@
...
+     *
+     * @param metricsEndpointUri The endpoint used to fetch metrics
+     * @return Metrics object
+     * @throws IOException If http request to retrieve Prometheus metrics failed
+     */
+    public static TestMetrics getPrometheusMetrics(URI metricsEndpointUri) throws IOException {
+        String response = HttpUtil.get(metricsEndpointUri, null, null, null, String.class);
+
+        TestMetrics metrics = new TestMetrics();
+        //Map<String, String> map = new LinkedHashMap<>();
```
mimaison, line 37: Do we need this line?

testsuite/common/src/main/java/io/strimzi/testsuite/oauth/common/TestUtil.java
```diff
@@ -0,0 +1,56 @@
...
+                    }
+                    if (inmatch) {
+                        result.add(line);
+                    }
+                }
+            }
+
+            return result;
+
+        } catch (Throwable e) {
```
mimaison, line 52: Again why are we catching `Throwable` and rethrowing as `RuntimeException`?

README.md
```diff
@@ -1109,6 +1117,179 @@ You may want to set these options globally as system properties or env vars to a
...
+
+The OAuth metrics also honor the `metric.reporters`, `metrics.num.samples`, `metrics.recording.level` and `metrics.sample.window.ms` configurations of Kafka runtimes. When OAuth metrics are enabled the OAuth layer creates duplicate instances of `JmxReporter` and the configured `MetricReporter`s since at the moment there is no other way to integrate with the existing metrics system of Kafka runtimes. 
+
+When OAuth metrics are enabled, managed beans are registered on demand, containing the attributes that are easily translated into Prometheus metrics.
+
+Each registered MBean contains two counter variables - `count`, and `timeTotal`.
+It also contains three gauge variables - `timeMin`, `timeMax` and `timeAvg`.
+
+The `count` contains the number of requests of the specific context (as represented by attributes).
+The `timeTotal` contains the cumulative time in milliseconds spent in the requests of the specific context.
```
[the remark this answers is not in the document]
mstruk: I think the only metrics in other Strimzi components is via OpenTelemetry integration, and I don't know how mappings work there. Kafka doesn't seem to include a unit in the metrics names. If I understand correctly the similar metrics there are expressed in nanoseconds. Some metrics that are in seconds have 'seconds' in the name.

I agree that adding `Ms` extension to time metrics would make it more explicit. We can do that.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
```diff
@@ -25,11 +29,13 @@ public class Config {
     public static final String OAUTH_ACCESS_TOKEN_IS_JWT = "oauth.access.token.is.jwt";
     public static final String OAUTH_CONNECT_TIMEOUT_SECONDS = "oauth.connect.timeout.seconds";
     public static final String OAUTH_READ_TIMEOUT_SECONDS = "oauth.read.timeout.seconds";
+    public static final String OAUTH_CONFIG_ID = "oauth.config.id";
+    public static final String OAUTH_ENABLE_METRICS = "oauth.enable.metrics";
```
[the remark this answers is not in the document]
mstruk: I don't know, we already have several boolean config options like: `oauth.check.issuer`, `oauth.check.access.token.type`, `oauth.check.audience`, `oauth.fail.fast`, `strimzi.authorizer.grant.when.no.delegate`, `strimzi.authorization.delegate.to.kafka.acl` ...

The common thread is that we name them using `imperative` mode rather than `passive`.
`oauth.enable.metrics` continues that theme, whereas `oauth.metrics.enable` breaks it in my opinion.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java
```diff
@@ -138,6 +173,21 @@ public boolean getValueAsBoolean(String key, boolean fallback) {
...
+     *
+     * @param key Config key
+     * @return Config value
+     */
+    public URI getValueAsURI(String key) {
+        String result = getValue(key);
+        try {
+            return URI.create(result);
+        } catch (Exception e) {
+            throw new RuntimeException("Bad URI value for key: " + key + ", value: " + result);
```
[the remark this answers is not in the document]
mstruk: Good catch.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
mstruk: This method is called as part of ScheduledExecutorService scheduled task which I don't want to stop working, which happens if the task method throws any exception. Additionally, I want the stacktrace to include the error message with some context info - the message set on the thrown RuntimeException.

I guess your proposal would be to do:
```
        } catch (Throwable ex) {
            addJwksHttpMetricErrorTime(ex, requestStartTime);
            if (ex instanceof Error) {
               throw (Error) ex;
            }
            throw new RuntimeException("Failed to fetch public keys needed to validate JWT signatures: " + keysUri, ex);
        }
```

But then there is `setupExecutorAndFetchInitialKeys()` method that rethrows the caught exception. There is some ergonomics in play when only RuntimeExceptions are possible that they don't have to be declared, so it just makes for a cleaner code IMHO to wrap anything here into RuntimeException as at the end of the day all is handled the same by my code higher up the stack.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
mstruk: Can you give an example where this would be really bad here?

In the case of any kind of failure here the authorization will be denied, and full stacktrace will be logged.
In this case a custom authorizer is actually not a separate piece without which the rest of the broker can function. If authorizer doesn't function, the broker doesn't function.

I guess by re-throwing an Error if it occurs that would signal Kafka to shutdown. Is that the desired behaviour in this case? Often times, especially on Kubernetes, it is easier to debug an issue without causing the JVM to shutdown.
But then what good is the system that is up, but not working as it should due to misconfiguration or limited resources.

testsuite/common/src/main/java/io/strimzi/testsuite/oauth/common/TestUtil.java
mimaison: Can we just catch `Exception`? We're not interested in doing anything with `Error` objects

mstruk: In this case I see no harm in doing it. Generally I want to add some context info or log all the failures in which case I would have to catch Throwable to prevent Errors from silently escaping through the catch harness.

testsuite/common/src/main/java/io/strimzi/testsuite/oauth/common/TestMetrics.java
mstruk: Removed.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/metrics/GrantsHttpMetricKeyProducer.java
mstruk: Yes, makes sense.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mstruk: Done.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mstruk: Done.

README.md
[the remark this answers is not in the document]
mstruk: Unfortunately the metrics names become quite unreadable when we add this. For example, the new names after JMX-to-Prometheus mapping become: timetotalms, timeavgms, timeminms, and timemaxms. Without 'ms' these are not perfect, but still well readable. With 'ms' it's just horrific.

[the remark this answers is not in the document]
mstruk: But addressing your idea of turning the names around, I personally don't find the result much better to be honest. But maybe let's ask more people what they think.

mstruk:
@tombentley @mimaison Is it time to merge?

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/MetricKey.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public Map<String, String> getAttrs() {
+        return attrs;
+    }
+
+    private String composeMBeanName(String name, Map<String, String> attrs) {
+        StringBuilder objName = new StringBuilder("strimzi.oauth:").append("type=").append(name);
+        Map<String, String> allAttrs = new HashMap<>(attrs);
+
+        for (String attr: SORTED_ATTRS) {
```
tombentley, line 67: A comment about why we need this special sorting semantics.

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/MetricKey.java
```diff
@@ -0,0 +1,101 @@
...
+        if (o == null || getClass() != o.getClass()) return false;
+        MetricKey metricKey = (MetricKey) o;
+        return Objects.equals(name, metricKey.name) && Objects.equals(attrs, metricKey.attrs);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, attrs);
+    }
+}
```
tombentley, line 101: Maybe add a `toString()` for debugging?

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/MetricKey.java
```diff
@@ -0,0 +1,101 @@
...
+        for (Map.Entry<String, String> ent: allAttrs.entrySet()) {
+            appendAttrToName(objName, ent.getKey(), ent.getValue());
+        }
+
+        return objName.toString();
+    }
+
+    private void appendAttrToName(StringBuilder objName, String attr, String val) {
+        objName.append(",").append(attr).append("=");
+        if (QUOTED_ATTRS.contains(attr)) {
```
tombentley, line 82: If attrs don't have to come from the `SORTED_ATTRS` then special casing these attrs for quoting seems a bit fragile.

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/JmxMetrics.java
```diff
@@ -0,0 +1,87 @@
...
+
+import javax.net.ssl.SSLException;
+import java.io.IOException;
+import java.net.ConnectException;
+import java.net.URI;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class JmxMetrics {
```
tombentley, line 17: 1. Javadoc
2. What's JMX about this class? It seems pretty independent of JMX to me?

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/AbstractMetricKeyProducer.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2017-2022, Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.oauth.metrics;
+
+import java.net.URI;
+
+public abstract class AbstractMetricKeyProducer implements MetricKeyProducer {
```
tombentley, line 9: This abstraction seems really weak. Is there really much value in a class holding two fields like this, or could they just be redeclared in the subclasses?

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/MetricKeyProducer.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2017-2022, Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.oauth.metrics;
+
+public interface MetricKeyProducer {
```
tombentley, line 7: Javadoc

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ConfigurationKey.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2017-2021, Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.oauth.services;
+
+import java.util.Objects;
+
+public class ConfigurationKey {
```
tombentley, line 9: Javadoc! What _is_ a ConfigurationKey?

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ValidatorKey.java
```diff
@@ -4,6 +4,8 @@
  */
 package io.strimzi.kafka.oauth.services;
 
+import io.strimzi.kafka.oauth.common.IOUtil;
+
 import java.util.Objects;
 
 public class ValidatorKey {
```
tombentley, line 11: What is a ValidatorKey?

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
```diff
@@ -59,10 +63,9 @@ public class JWTSignatureValidator implements TokenValidator {
 
     private static final Logger log = LoggerFactory.getLogger(JWTSignatureValidator.class);
 
-    private final DefaultJWSVerifierFactory verifierFactory = new DefaultJWSVerifierFactory();
```
tombentley, line 62: I realise this boat has sailed, but why is this class shouty `JWT...`, but we've got `JmxMetrics` etc?

examples/kubernetes/kafka-oauth-single-authz.yaml
```diff
@@ -16,6 +16,8 @@ spec:
           validIssuerUri: http://keycloak:8080/auth/realms/kafka-authz
           jwksEndpointUri: http://keycloak:8080/auth/realms/kafka-authz/protocol/openid-connect/certs
           userNameClaim: preferred_username
+          checkAccessTokenType: false
+          customClaimCheck: "@.typ && @.typ == 'Bearer'"
```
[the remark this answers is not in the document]
mstruk: No, it's unrelated. Maybe I should put it in another PR?

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/MetricKey.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+public class MetricKey {
```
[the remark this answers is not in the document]
mstruk: It would actually be better named SensorKey, but that will require renaming a whole bunch of classes to keep it consistent.

Do we have some policy on when to add JavaDoc and when not? Not all public classes should be considered a public API do we have a policy to mark documented classes as not considered a public API?

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/MetricKey.java
```diff
@@ -0,0 +1,101 @@
...
+    final Map<String, String> attrs;
+    final String mbeanName;
+
+    public MetricKey(String name, Map<String, String> attrs) {
+        this.name = name;
+        this.attrs = Collections.unmodifiableMap(attrs);
+        this.mbeanName = composeMBeanName(name, attrs);
+    }
+
+    public static MetricKey of(String name, Map<String, String> attrs) {
```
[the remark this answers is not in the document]
mstruk: Yup, I noticed that too.

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/MetricKey.java
mstruk: Actually, as a result of the transition to Kafka Metrics, this code, while producing a valid MBean ObjectName, is not used for the ObjectName any more, just to produce a unique Sensor key that incorporates a base name and all the attributes. We could concatenate a name and a string representation of the attributes map, and it would be a unique key just as well. I'll clean this up.

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/MetricKey.java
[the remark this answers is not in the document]
tombentley: > that will require renaming a whole bunch of classes to keep it consistent.

But your IDE makes refactoring really easy, so if there are better names I don't understand why such renaming is a barrier.

> Do we have some policy on when to add JavaDoc and when not?

I don't see the purpose of Javadoc being always documentation for public consumption. It's a really good way of recording what the purpose of the class is, what invariants/assumptions/guarantees it has/makes etc. The ultimate user of this documentation will be someone maintaining this code in the future (most likely you), and anyone else who wants to understand how the classes fit together (me, or other reviewers, for example).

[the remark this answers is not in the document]
mstruk: > But your IDE makes refactoring really easy, so if there are better names I don't understand why such renaming is a barrier.

Never mind, it wasn't meant that way.

> I don't see the purpose of Javadoc being always documentation for public consumption. It's a really good way of recording what the purpose of the class is, what invariants/assumptions/guarantees it has/makes etc. The ultimate user of this documentation will be someone maintaining this code in the future (most likely you), and anyone else who wants to understand how the classes fit together (me, or other reviewers, for example).

I'll add more JavaDoc, and very nice you pointed out what exactly you would like to be explained.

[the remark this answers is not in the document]
tombentley: Thanks!

examples/kubernetes/kafka-oauth-single-authz.yaml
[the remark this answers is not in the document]
tombentley: I guess it's not a problem to keep it here.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
mstruk: Sometimes the shouty version is more readable so I opt for a shouty one.
It would be ideal to have consistency, but then readability is also a factor. Granted it's a subjective criteria, but then a lot of software development is choosing between multiple approaches and weighing pros and cons.

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/JmxMetrics.java
mstruk: True, after refactoring all the JMX parts were removed, but the class name remained.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ValidatorKey.java
mstruk: Added javadoc.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/ConfigurationKey.java
mstruk: Added javadoc.

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/MetricKeyProducer.java
mstruk: Added javadoc.

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/MetricKey.java
mstruk: Added toString()

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/JmxMetrics.java
mstruk: Added javadoc, renamed class to MetricsUtil.

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/MetricKey.java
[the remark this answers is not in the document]
mstruk: Added javadoc. Renamed class to SensorKey.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
```diff
@@ -0,0 +1,186 @@
...
+
+    private final Map<MetricKey, Sensor> sensorMap = new ConcurrentHashMap<>();
+
+    OAuthMetrics(Map<String, ?> configMap) {
+        this.configMap = configMap;
+        this.config = new Config(configMap);
+        this.metrics = initKafkaMetrics();
+    }
+
+    public void addTime(MetricKey key, long time) {
```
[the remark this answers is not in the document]
mstruk: Added javadoc, renamed parameter to timeMs. Left addTime name as-is since IDE helps with parameter names so it's easy to figure out that it's millis.

testsuite/common/src/main/java/io/strimzi/testsuite/oauth/common/TestUtil.java
mstruk: Continued in #147 .

mimaison (approved):

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/SensorKey.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Create a new instance of SensorKey from name and a map of attributes
+     *
+     * @param name The metric name
+     * @param attributes The map of attributes
+     * @return The new instance of SensorKey
+     */
+    public static SensorKey of(String name, Map<String, String> attributes) {
+        return new SensorKey(name, attributes);
```
tombentley, line 61: If `attributes` is a `HashMap` then the order of the attributes will depend on the size of the array backing the hashmap, and therefore two `SensorKey` instances with the same (`equals`) `attributes` could have different `id`s. Similarly you could create one instance via this factory and another via the `varargs` factory and the same attributes would result in different ids. Perhaps the simplest thing would be to use a `TreeMap` in the varargs `of` and copy the `attributes` here into a `TreeMap` guaranteeing the iteration order, which should be enough to ensure the map's `toString` is also stable in the face of undefined map orderings.

tombentley (commented):
@mstruk those docs are excellent and exactly the sort of thing I was looking for: Documenting the intent behind the classes. Thanks! Just only last comment this time.

oauth-common/src/main/java/io/strimzi/kafka/oauth/metrics/SensorKey.java
mstruk: That's a fair point. Currently things have been working fine because all the usages are passing attributes as LinkedHashMap that contains a proper ordering. For the expected MBean name construction the order of the attributes is also important. I made a simple solution that should address both predictable `id` generation and expected attribute ordering for the MBean names.

tombentley (approved):
LGTM, thanks @mstruk

Issue #124, opened 2021-09-09
hydra-import Dockerfile in testsuite contains hardcoded reference to hydra binary


rposts:
I was trying to run integration tests on `s390x` architecture and noticed that `hydra-test` failed.  After some investigation I found that this was caused due to incompatible `strimzi-oauth-testsuite/hydra-import` image generated by the build step.

It turns out that `hydra-import` [Dockerfile](https://github.com/strimzi/strimzi-kafka-oauth/blob/main/testsuite/docker/hydra-import/Dockerfile) downloads `x86` version of [hydra_1.0.0_Linux_32-bit.tar.gz](https://github.com/ory/hydra/releases/download/v1.0.0/hydra_1.0.0_Linux_32-bit.tar.gz) which causes problems on `s390x` architecture.

I made following changes to make it work on `s390x` arch and think it may also work on `x86` arch. All I am doing here is to compile `hydra` from source instead of downloading a specific version:
```diff
index 6d6e0f2..8de6f6e 100644
--- a/testsuite/docker/hydra-import/Dockerfile
+++ b/testsuite/docker/hydra-import/Dockerfile
@@ -1,10 +1,15 @@
 FROM alpine

-RUN apk add -U --no-cache bash ca-certificates curl \
-    && mkdir /tmp/hydra \
-    && cd /tmp/hydra \
-    && wget -q https://github.com/ory/hydra/releases/download/v1.0.0/hydra_1.0.0_Linux_32-bit.tar.gz \
-    && tar xvzf hydra_1.0.0_Linux_32-bit.tar.gz \
-    && cp hydra /usr/bin/ \
+COPY --from=golang:1.16-alpine /usr/local/go/ /usr/local/go/
+ENV PATH="/usr/local/go/bin:${PATH}"
+
+RUN apk add -U --no-cache bash ca-certificates curl git make gcc musl-dev \
+    && cd /tmp \
+    && export GOPATH=/opt/go/ \
+    && git clone -b v1.8.5 https://github.com/ory/hydra.git \
+    && cd hydra \
+    && GO111MODULE=on make install-stable \
+    && cp $GOPATH/bin/hydra /usr/bin/ \
     && cd \
     && rm -rf /tmp/hydra
+ENV GODEBUG=x509ignoreCN=0
```
Please let me know if this is an acceptable solution and whether `Dockerfile` can be changed accordingly.

Thanks.

scholzj:
[...]

PR #144, opened 2022-04-12
Add s390x support to hydra testcase

rposts:
This PR address changes required to run `hydra` testcase on `s390x` architecture.

As per https://github.com/strimzi/strimzi-kafka-oauth/issues/124, a `s390x` maven profile is added to `hydra-import` testcase which will only be triggered on `s390x` architecture to build necessary  images.

@mstruk FYI


mstruk:
While the addition is not part of the main build (deployable jars), it is a part of the testsuite, which runs multiple times on every CI run.

Rather than pulling into the build the already built go-lang binary, the build of the binary from go-lang sources is performed as a part of the testsuite run, which probably makes this take quite some time.
If this time is short (arbitrary number, but let's say 30s max) then I don't really see much of a problem.
But make sure that the build is only performed if the image does not yet exist in the Docker container. The testsuite in CI runs multiple times with multiple Kafka versions. You don't want to perform the same build, producing the same binary again and again.

While I agree that this build doesn't quite belong there, it has to be somewhere, and if it works for @rposts I don't really object, as long as it doesn't affect any other platform (which based on the activation of the profile should be the case).

Ideally that part would be outside this project, but then - where?
- https://github.com/strimzi/hydra ? 
- https://github.com/rposts/hydra ?  
- here, under testsuite but not part of the testsuite run itself

And the binary of the build should then be uploaded somewhere to quay.io or hub.docker.com, where it's publicly available and should have a version that reflects the architecture in order to prevent confusion with default architectures (I guess). In that case our script would have to fetch the s390x version and re-tag it locally for the dependency to be met locally.

The current solution is convenient, but if the build takes a longer time the CI may terminate the testsuite run before it has a chance to complete, especially when testsuite run uses some older kafka version.

See: https://github.com/strimzi/strimzi-kafka-oauth/blob/main/.travis/build.sh for the actual script the CI executes on every run to get a better idea.

But as I said, I can not test this, and as long as it doesn't affect anything else, and works for you I'd say it's fine with me.

scholzj:
I don't think you really need a separate repo. And we definitely don't want t build and push it under Strimzi. But I think the logic can be easily added for s390x runs into the `./travis` directory to the other build scripts. It seems really weird to me to have it as part of the Maven build when it is just something for one platform which almost nobody uses.

mstruk:
I see what you mean now.

Yeah, since the addition really has nothing to do with Java why even use maven to build it when the same can be done by using `docker build` command directly in a new script (let's call it `.travis/prepare-s390x.sh`) which can be invoked at the start of `build.sh` and would only execute if it determines that the build is running on s390x platform.

Let's do it like that then.

mstruk (changes requested):
Please modify according to the comments above.

rposts:
@mstruk - I have made some changes to the Travis build script where hydra image is only built for `s390x` arch when running the testcases.  Currently, `s390x` arch does not have `quay.io/strimzi/kafka:0.29.0-kafka-3.1.0` image available but it is expected to be there in the near future. 
I also think appropriate modifications will be needed in `.travis.yml` file to enable testing on `s390x`.  Let's see if this brings us closer to addressing the original issue.
Let me know your thoughts.
Thanks!

.travis/build.sh
```diff
@@ -38,44 +38,61 @@ mvn spotbugs:check
 # Run testsuite with java 8 only
 if [ ${JAVA_MAJOR_VERSION} -eq 1 ] ; then
 
-  docker pull oryd/hydra:v1.8.5
+  arch=$(uname -i)
```
mstruk, line 41: `uname -i` doesn't work on MacOS, breaking the build.

```
uname: illegal option -- i
usage: uname [-amnprsv]
```

mstruk (changes requested):
Generally it looks good, but `-i` option of `uname` doesn't work on MacOS.

.travis/build.sh
rposts: Changed to `uname -m` - this should work.

mstruk: Looks good to me.

mstruk (approved):

.travis/build.sh
```diff
@@ -38,44 +38,61 @@ mvn spotbugs:check
...
-
-  set -e
+  if [ "$arch" == 's390x' ]; then
+    # Build s390x compatible hydra images
+    docker build --target hydra-import -t strimzi-oauth-testsuite/hydra-import:latest -f ./testsuite/docker/hydra-import/Dockerfile.s390x .
+    docker build --target oryd-hydra -t oryd/hydra:v1.8.5 -f ./testsuite/docker/hydra-import/Dockerfile.s390x .
+    mvn test-compile spotbugs:check -e -V -B -f testsuite
+    set +e
+    clearDockerEnv
+    docker pull quay.io/strimzi/kafka:0.29.0-kafka-3.1.0
```
scholzj, line 52: This image does not exist. So how come this is not failing?

rposts: @scholzj - this was tested in my local dev environment where I had manually built the images.  For this reason testing is not yet enabled in `.travis.yml` for `s390x` platform.

scholzj: Ahh, ok. That makes sense. So I guess we should wait for the 0.29.0 release before merging this, so that we can enable it and test it? Hopefully, 0.29 will be out soon. We are waiting only for Kafka 3.1.1.

rposts: During my testing I had made similar changes to test:
```
set +e
clearDockerEnv
docker pull quay.io/strimzi/latest-kafka-3.1.0
docker tag quay.io/strimzi/latest-kafka-3.1.0 quay.io/strimzi/kafka:0.28.0-kafka-3.1.0
#docker pull quay.io/strimzi/kafka:0.28.0-kafka-3.1.0
mvn -e -V -B clean install -f testsuite -Pkafka-3_1_0
EXIT=$?
exitIfError
```

rposts: Yes - let's wait until all images are available and initiate a test run.

Thanks @scholzj .

scholzj: 0.29.0 is out now. So you should be able to try to update to it.

rposts: Thanks @scholzj - I will test it shortly.

rposts: @scholzj I tested it with 0.29.0 but ran into an issue where [testsuite/docker/hydra-import](https://github.com/strimzi/strimzi-kafka-oauth/blob/main/testsuite/docker/hydra-import/pom.xml) fails because docker plugin tries to build image using amd `Dockerfile`.  I am not sure if there is a way around it without bringing in `s390x` profile back as implemented earlier. 

Here is [TravisCI](https://app.travis-ci.com/github/linux-on-ibm-z/strimzi-kafka-oauth/jobs/571558217)  output which shows amd64 hydra image being built instead of `s390x`.

Let me know your thoughts.

Corresponding changes in `.travis/build.sh`:
```
  if [ "$arch" == 's390x' ]; then
    # Build s390x compatible hydra images
    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib/s390x-linux-gnu/jni
    docker build --target hydra-import -t strimzi-oauth-testsuite/hydra-import:latest -f ./testsuite/docker/hydra-import/Dockerfile.s390x .
    docker build --target oryd-hydra -t oryd/hydra:v1.8.5 -f ./testsuite/docker/hydra-import/Dockerfile.s390x .
    set +e
    clearDockerEnv
    docker pull quay.io/strimzi/kafka:0.29.0-kafka-3.1.1
    mvn -e -V -B clean install -f testsuite -Pcustom -Dkafka.docker.image=quay.io/strimzi/kafka:0.29.0-kafka-3.1.1
    EXIT=$?
    exitIfError
    set -e
  else
```

scholzj: I have no idea how the s390x environment works on Travis. Does the Docker engine run on s390x as well? If it can build amd64 hydra, can't it just run the amd64 hydra image in the first place? Then we would not need to build anything.

rposts: My understanding is that TravisCI on `s390x` is LXD based and build jobs run within LXD container. Using amd64 hydra binary included in the hydra image will cause `exec format error` due to architectural differences. This particular problem only occurs for a specific testcase `hydra-import` of `docker` module testsuite.  Everything works (s390x specific) for `hydra-test` module.

scholzj: Maybe that needs to be moved into some shell script as well?

mstruk: I'll see if I can use plain `curl` in `hydra-import` so there is no need for installing 'hydra' executable in it. That would probably fix the issue. It would still use the 'alpine' image from Docker Hub though.

rposts: Thanks @mstruk - I don't fully understand the code but `alpine` image should work on s390x as it is multi-arch enabled. I will be happy to test your changes on this platform.

If this works then changes in `.travis/build.sh` to build `oryd-hydra` image for `s390x` is all that is needed to address this issue.

mstruk: See https://github.com/linux-on-ibm-z/strimzi-kafka-oauth/pull/1

rposts: @mstruk - looks good. 

Unfortunately, testcases continue to fail due to use of `Arquiliian Cube` extension which fails to interact with docker on `s390x`. I have tested a patch which compiles and installs s390x compatible `arquillian` jars in the local maven repo. This enables docker communication with `arquillian` on s390x.

For the purpose of this PR I think it is acceptable to have `hydra` issue addressed. I am away next week but will take this upon my return. Your changes above to remove hydra binary along with my changes to build `oryd:hydra` should take care of things.

Thanks!

mstruk:
@rposts Do I understand correctly that the testsuite doesn't really work on s390x despite the changes you made, because you can not get Arquillian to work?

That would probably make this PR pointless, since the changes only come into play when testsuite is built and run?

rposts:
@mstruk - `<arquillian.cube.version>1.18.3-SNAPSHOT</arquillian.cube.version>` (as opposed to existing `<arquillian.cube.version>1.18.2</arquillian.cube.version>` version in ../testsuite/pom.xml) works for `s390x` and connects with Docker but strangely it is unable to detect if Kafka server has started in TravisCI environment:

 ```INFO: docker exec sh -c grep 'started (kafka.server.KafkaServer)' < /tmp/logs/server.log, exit status:1, stdout:``` [Travis CI link](https://app.travis-ci.com/github/linux-on-ibm-z/strimzi-kafka-oauth/jobs/575669327)

I do not have this issue in my local environment.

This is turning out to be more complicated than I earlier thought.  However, I still think it will be useful to have `s390x` changes in place so that any future debugging can be made simpler.

mstruk:
@rposts Very well. Let's merge it then.
@scholzj WDYT?

scholzj:
I guess first of all this needs to be rebased and we need to see the tests passing.

scholzj:
@rposts Can you please rebase this so that the tests are triggered? Thanks.

rposts:
Hi @scholzj - I will get to it shortly. Thanks!

scholzj (approved):

scholzj:
Thanks @rposts

PR #147, opened 2022-05-17
Error handling

mstruk:
Review the handling of exceptions, particularly the handling of Errors within `catch (Throwable t)` blocks.
Also update SpotBugs to latest version and address all the warnings it issues.

This PR depends on #141.

### DIFF
--- a/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
+++ b/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
@@ -388,7 +388,11 @@ public class JWTSignatureValidator implements TokenValidator {
                     throw new TokenValidationException("Token validation failed: The signing key is no longer valid (kid:" + kid + ")");
                 } else {
                     // Request quick keys refresh
-                    fastScheduler.scheduleTask();
+                    try {
+                        fastScheduler.scheduleTask();
+                    } catch (RuntimeException e) {
+                        log.error("Failed to reschedule JWKS keys refresh: ", e);
+                    }
                     throw new TokenValidationException("Token validation failed: Unknown signing key (kid:" + kid + ")");
                 }
             }
--- a/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
+++ b/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
@@ -23,6 +23,7 @@ import io.strimzi.kafka.oauth.jsonpath.JsonPathQuery;
 import io.strimzi.kafka.oauth.metrics.JwksHttpSensorKeyProducer;
 import io.strimzi.kafka.oauth.metrics.SensorKeyProducer;
 import io.strimzi.kafka.oauth.services.OAuthMetrics;
+import io.strimzi.kafka.oauth.services.ServiceException;
 import io.strimzi.kafka.oauth.services.Services;
 import org.apache.kafka.common.utils.Time;
 import org.slf4j.Logger;
@@ -362,7 +363,7 @@ public class JWTSignatureValidator implements TokenValidator {
 
         } catch (Throwable ex) {
             addJwksHttpMetricErrorTime(ex, requestStartTime);
-            throw new RuntimeException("Failed to fetch public keys needed to validate JWT signatures: " + keysUri, ex);
+            throw new ServiceException("Failed to fetch public keys needed to validate JWT signatures: " + keysUri, ex);
         }
     }
 
@@ -443,7 +444,7 @@ public class JWTSignatureValidator implements TokenValidator {
             principal = principalExtractor.getSub(tokenJson);
         }
         if (principal == null) {
-            throw new RuntimeException("Failed to extract principal - check usernameClaim, fallbackUsernameClaim configuration");
+            throw new ValidationException("Failed to extract principal - check usernameClaim, fallbackUsernameClaim configuration");
         }
         return principal;
     }
--- a/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
+++ b/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
@@ -5,6 +5,7 @@
 package io.strimzi.kafka.oauth.validator;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.BooleanNode;
 import io.strimzi.kafka.oauth.common.JSONUtil;
 import io.strimzi.kafka.oauth.common.PrincipalExtractor;
 import io.strimzi.kafka.oauth.common.TimeUtil;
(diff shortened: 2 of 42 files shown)


mstruk:
I updated the SpotBugs version and it started issuing some new warnings. Specifically the one that says it's bad practice to `throw RuntimeException()` in your code (rule THROWS_METHOD_THROWS_RUNTIMEEXCEPTION).

I addressed some of the warnings by introducing custom extensions of RuntimeException. But it's clear that I'm basically doing it to silence the warnings. I see no real value in these exceptions and they are quite unnecessary.
In some places I explicitly silenced the rule as I find introducing additional runtime exceptions plain harmful there. 

The blanket rule for throwing the RuntimeException would best be turned off IMHO.
But for now I indulged it.

scholzj:
@mstruk In the operators repo, we have an exclude file for disabling the rules we _don't like_: https://github.com/strimzi/strimzi-kafka-operator/blob/main/.spotbugs/spotbugs-exclude.xml

But I guess someone like @tombentley would be better to comment on how useful a specific rule is. I personally would agree that I don't see too much value from a rule like this.

mstruk:
@mimaison Any comments?
@tombentley Any thoughts on this?

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
```diff
@@ -384,7 +385,11 @@ public TokenInfo validate(String token) {
                     throw new TokenValidationException("Token validation failed: The signing key is no longer valid (kid:" + kid + ")");
                 } else {
                     // Request quick keys refresh
-                    fastScheduler.scheduleTask();
+                    try {
+                        fastScheduler.scheduleTask();
+                    } catch (RuntimeException e) {
```
mimaison, line 390: Should we catch `ServiceException` here?

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
```diff
@@ -272,14 +272,14 @@ public TokenInfo validate(String token) {
...
-            throw new RuntimeException("Failed to introspect token - send, fetch or parse failed: ", e);
+            throw new ValidationException("Failed to introspect token - send, fetch or parse failed: ", e);
         }
 
         boolean active;
         try {
             active = response.get("active").asBoolean();
         } catch (Exception e) {
-            throw new RuntimeException("Failed to introspect token - invalid response: \"active\" attribute is missing or not a boolean (" + response.get("active") + ")");
+            throw new ValidationException("Failed to introspect token - invalid response: \"active\" attribute is missing or not a boolean (" + response.get("active") + ")");
```
mimaison, line 282: Looking at the Javadocs `get()` and `asBoolean()` don't seem to throw but instead return `null` and `false`. Should we check the return value of `get()` instead?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
```diff
@@ -598,7 +604,7 @@ private JsonNode fetchAuthorizationGrants(String token) {
             throw e;
         } catch (Exception e) {
             addGrantsHttpMetricErrorTime(e, startTime);
-            throw new RuntimeException("Failed to fetch authorization data from authorization server: ", e);
+            throw new ServiceException("Failed to fetch authorization data from authorization server: ", e);
```
mimaison, line 607: Could we get rd of the `catch (HttpException e)` block above?

mimaison (commented):
Thanks for the PR, I left a few comments.

pom.xml
```diff
@@ -286,7 +286,7 @@
             <plugin>
                 <groupId>com.github.spotbugs</groupId>
                 <artifactId>spotbugs-maven-plugin</artifactId>
-                <version>3.1.12</version><dependencies>
+                <version>4.6.0.0</version><dependencies>
```
tombentley, line 289: `<dependencies>` on a new line?

tombentley (commented):
Thanks, this is a nice improvement @mstruk. Is there any value in a single top-level `OauthException`, rather than having multiple exception types extend `RuntimeException`?

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
mstruk: No. `ServiceException` extends `RuntimeException`, and I prefer to catch broadly here. The task scheduling error should be logged but should not have any other side effects as that would result in a different error message that is returned to the client doing the authenticating.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
mstruk: You mean, if the JSON is:
```
{
    "active": "True"
}
```
the code will currently actually set `active` to `false`, because only string with case-sensitive value of "true" will be converted to the boolean value of `true`.

That's a good point. While authorization servers should return a boolean attribute we should maybe be more strict, and we currently don't really detect the 'not a boolean' case.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
mstruk: No. The exception handling in several places up the stack is different when HttpException occurs.

mstruk:
> Thanks, this is a nice improvement @mstruk. Is there any value in a single top-level `OauthException`, rather than having multiple exception types extend `RuntimeException`?

I considered it but it seemed just another layer, more for style than for anything useful.

tombentley (approved):

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
mstruk: Fixed.

pom.xml
mstruk: Fixed.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
```diff
@@ -272,16 +273,14 @@ public TokenInfo validate(String token) {
...
         }
 
-        boolean active;
-        try {
-            active = response.get("active").asBoolean();
-        } catch (Exception e) {
-            throw new RuntimeException("Failed to introspect token - invalid response: \"active\" attribute is missing or not a boolean (" + response.get("active") + ")");
+        JsonNode activeAttr = response.get("active");
+        if (!(activeAttr instanceof BooleanNode)) {
+            throw new ValidationException("Failed to introspect token - invalid response: \"active\" attribute is missing or not a boolean (" + activeAttr + ")");
```
mimaison, line 281: Maybe we could log the response too, WDYT?

mstruk: Not sure about that. My concern is that some info that should not enter the log files may be part of the response.

The root of the problem will ultimately be a misconfigured or incompatible authorization server. But if by any chance it includes the raw token in the response and we log it, it's possible that the token is actually valid, but the response contains something like "valid": "true" rather than "active": true. We would log it as error, and if it contains the access token (it's not supposed to, but if the server is some bespoke in-house thing all is possible) the log collection may insert it into some central log system, and there you have it in the clear a valid access token that is possibly usable on some other Kafka instance (maybe one configured with JWT signature check, rather than with introspection validation).

mimaison (approved):

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
mstruk: Actually you can log all HTTP responses already by setting `io.strimzi.kafka.oauth.common.HttpUtil` Logger level to `TRACE`.
For development that should be enough already to get to the bottom of the problem here.

PR #149, opened 2022-05-26
Add `oauth.jwks.ignore.key.use` config option

mstruk:
Set this option to `true` in order to use all the keys in the JWKS response for token signature validation, regardless of their `use` attribute.
This makes it possible to use authorization servers that don't specify `use` attribute in JWKS keys.

### DIFF
--- a/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
+++ b/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
@@ -107,6 +107,7 @@ public class JWTSignatureValidator implements TokenValidator {
      * @param refreshSeconds The optional time interval between two consecutive regular JWKS keys refresh runs
      * @param refreshMinPauseSeconds The optional minimum pause between two consecutive JWKS keys refreshes.
      * @param expirySeconds The maximum time to trust the unrefreshed JWKS keys. If keys are not successfully refreshed within this time, the validation will start failing.
+     * @param ignoreKeyUse Should any key present in JWKS key set be considered a public key for signature checking
      * @param checkAccessTokenType Should the 'typ' claim in the token be validated (be equal to 'Bearer')
      * @param audience The optional audience
      * @param customClaimCheck The optional JSONPath filter query for additional custom claim checking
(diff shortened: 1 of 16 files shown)


oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
```diff
@@ -113,6 +114,7 @@ public class JWTSignatureValidator implements TokenValidator {
      * @param readTimeoutSeconds The maximum time to wait for response from authorization server after connection has been established and request sent (in seconds)
      * @param enableMetrics The switch that enables metrics collection
      * @param failFast Should exception be thrown during initialisation if unable to retrieve JWKS keys
+     * @param ignoreKeyUse Should any key present in JWKS key set be considered a public key for signature checking
```
scholzj, line 117: Should we keep these in the right order in which they are specified in the method?

testsuite/README.md
```diff
@@ -181,6 +187,6 @@ Thus, you don't need to use the latest local build of strimzi/kafka libraries to
 
 But if you want you can specify the kafka image to use for the test as follows:
 
-    mvn clean test -Dkafka.docker.image=strimzi/kafka:build-kafka-2.6.0  -f testsuite/keycloak-auth-tests
+    mvn clean test -Dkafka.docker.image=quay.io/strimzi/kafka:0.28.0-kafka-3.1.0 -f testsuite/keycloak-auth-tests
```
scholzj, line 190: Strimzi 0.29.0 supports Kafka 3.2.0. (just saying, no need to update it right here in this PR)

scholzj (approved):
Some nits. But LGTM otherwise.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/JWTSignatureValidator.java
mstruk: Good catch. Strange that neither IDEA nor SpotBugs raised any warnings for this.

testsuite/README.md
mstruk: Yeah, another PR to add a profile for 3.2.0 and make it the default.

PR #151, opened 2022-07-02
Update oauth.sh

buuhsmead:
--insecure option was not used in curl command

Signed-off-by: buuhsmead <buuhsmead@gmail.com>


mstruk (approved):
Makes sense to also support insecure https when `--insecure` option is used.

scholzj (approved):

scholzj:
Thanks for the PR!

PR #153, opened 2022-07-14
Unprotected truststores

mstruk:
Add support for truststores that don't have a password set.

### DIFF
--- a/oauth-common/src/test/java/io/strimzi/kafka/oauth/common/CommonTests.java
+++ b/oauth-common/src/test/java/io/strimzi/kafka/oauth/common/CommonTests.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2017-2022, Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.oauth.common;
+
+import org.junit.Test;
+
+public class CommonTests {
+
+    @Test
+    public void doTest() throws Exception {
+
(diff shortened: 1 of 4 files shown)


mstruk:
@scholzj This one is ready for review. Whenever you find some time ...

oauth-common/src/test/java/io/strimzi/kafka/oauth/common/CommonTests.java
```diff
@@ -0,0 +1,39 @@
...
+        new UnprotectedTruststoreTest().doTest();
+    }
+
+    private void logStart(String msg) {
+        System.out.println();
+        System.out.println();
+        System.out.println("========    "  + msg);
+        System.out.println();
+        System.out.println();
+    }
```
scholzj, line 38: Is this intentionally left here?

scholzj (approved):
One nit, but LGTM otherwise. Thanks.

oauth-common/src/test/java/io/strimzi/kafka/oauth/common/CommonTests.java
mstruk: It is. I use it in other tests as well. It removes any doubts about logging being turned off and allows following progress with minimal output. To reduce the time of starting the dependency containers, many tests are merged into one single JUnit test, and the output helps with following the progress as the 'subtests' within the single JUnit test are started.

scholzj: Ok, fine with me in that case.

PR #154, opened 2022-07-22
Update jackson dep versions to 2.12.6*

grdryn:
I'm not sure what the policy for upgrading this here is normally, but this upgrades to match the versions specified in Apache Kafka 3.2.0:

https://github.com/apache/kafka/blob/3.2.0/gradle/dependencies.gradle#L69..L70

Signed-off-by: Gerard Ryan <gryan@redhat.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,8 +81,8 @@
         <sonatype.nexus.staging>1.6.3</sonatype.nexus.staging>
 
         <kafka.version>3.2.0</kafka.version>
-        <jackson.version>2.10.5</jackson.version>
-        <jackson.databind.version>2.10.5.1</jackson.databind.version>
+        <jackson.version>2.12.6</jackson.version>
+        <jackson.databind.version>2.12.6.1</jackson.databind.version>
         <jsonpath.version>2.6.0</jsonpath.version>
         <junit.version>4.13.1</junit.version>
         <slf4j.version>1.7.26</slf4j.version>


scholzj (approved):
@grdryn I think it is fine to upgrade it if the test pass. 

Note: But in general, I think that most of Jackson is excluded (at least when adding it to the Kafka containers in Strimzi operators repo), so the version listed here is anyway not used in the containers for example.

grdryn:
@mstruk bumping this in case you missed it :slightly_smiling_face: 

I'm not 100% sure what's going on with the failing Travis check...it looks like it might have been something like an intermittent infra issue or a build flake or something, but I don't have the ability to re-trigger to check if it would pass now.

scholzj:
I restarted the build and it seemed to pass.

mstruk:
> @mstruk bumping this in case you missed it 🙂

Sorry for no feedback. I usually get on to bumping versions after other open PRs are merged.
And there's a general round of dependencies reviews before doing a release.

At the end of the day, the server side OAuth libraries have to use whatever is packaged with Kafka, or vice versa - if you want a different version Kafka has to work with that so it's a matter of packaging the different version with Kafka.

The testsuite performs the tests using strimzi Kafka images with 1.0.0-SNAPSHOT versions of OAuth jars deployed to the CLASSPATH in front of the rest of the CLASSPATH. Thus we only test compile time pass with these dependencies versions for Kafka broker usage. 

The client side version used in tests is the one resolved by maven though. So at least the client side should be tested runtime.

I'll check the failing CI to see what could be the problem.

PR #159, opened 2022-09-29
Update dependencies and release notes

mstruk:


examples/docker/spring/pom.xml
```diff
@@ -51,7 +51,7 @@
             <dependency>
                 <groupId>org.yaml</groupId>
                 <artifactId>snakeyaml</artifactId>
-                <version>1.31</version>
+                <version>1.32</version>
```
scholzj, line 54: 1.33 is out. In case you want to use that. (just saying, up to you I guess if you want 1.33 or stay with 1.32)

pom.xml
```diff
@@ -80,7 +80,7 @@
         <spotbugs.version>4.7.0</spotbugs.version>
         <sonatype.nexus.staging>1.6.3</sonatype.nexus.staging>
 
-        <kafka.version>3.2.1</kafka.version>
+        <kafka.version>3.2.3</kafka.version>
         <jackson.version>2.12.6</jackson.version>
         <jackson.databind.version>2.12.6.1</jackson.databind.version>
```
scholzj, line 85: Kafka 3.2.3 and 3.3.0 uses JAckson 2.13 IIRC. Would it make sense to align on that?

scholzj (approved):
Some nits for consideration. Feel free to ignore them if you think it is better to stick with what you have now.

pom.xml
mstruk: Missed the updates. Will update.

examples/docker/spring/pom.xml
mstruk: Let's update then.

PR #161, opened 2022-10-03
Fix keycloak example + improve CI build to catch a similar issue next time

mstruk:


examples/docker/keycloak/compose-ssl.yml
```diff
@@ -14,4 +14,4 @@ services:
       KEYCLOAK_PASSWORD: "admin"
       KEYCLOAK_HTTPS_PORT: 8443
       PROXY_ADDRESS_FORWARDING: "true"
-    command: "-Dkeycloak.profile.feature.upload_scripts=enabled"
+    command: "-Dkeycloak.profile.feature.upload_scripts=enabled -b 0.0.0.0"
```
see-quick, line 17: Shouldn't be `0.0.0.0` IP address the default one?

mstruk: Apparently the default one is some network interface with IP that is not accessible from other docker instances.

PR #180, opened 2023-01-05
Minimise initial grants requests for 'same token, different session' case

mstruk:
If grants for the token exist in another session, reuse them, rather than fetch them again. Also introduce per-access-token semaphore to prevent parallel fetching of grants for the same token.

Add `strimzi.authorization.reuse.grants` config option to enable this optimisation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -700,6 +700,17 @@ since they are not yet available. The following option enables reattempting the 
 receive the `AuthorizationException`. The default value is '0', meaning 'no retries'. Provide the value greater than '0' to set the number of repeated attempts:
 - `strimzi.authorization.http.retries` (e.g.: "1" - if initial fetching of grants for the session fails, immediately retry one more time)
 
+A single client typically uses a single unique access token for the concurrent sessions to the Kafka broker.
+As a result, the number of active tokens on the broker is generally less than the number of active sessions (connections).
+ However, keep in mind that this is replicated across all Kafka brokers in the cluster, as clients maintain active sessions to multiple brokers.
+
+New sessions will, by default, request the latest grants from the Keycloak in order for any changes in permissions to be reflected immediately.
+You can change this, and reuse the grants for the token, if they have previously been fetched due to the same token already having been used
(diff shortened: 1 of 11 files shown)


mstruk:
The `ppc64le` CI build keeps having issues. The testsuite run (linux) passed.

README.md
```diff
@@ -700,6 +700,17 @@ since they are not yet available. The following option enables reattempting the
 receive the `AuthorizationException`. The default value is '0', meaning 'no retries'. Provide the value greater than '0' to set the number of repeated attempts:
 - `strimzi.authorization.http.retries` (e.g.: "1" - if initial fetching of grants for the session fails, immediately retry one more time)
 
+A single client typically uses a single unique access token for the concurrent sessions to the Kafka broker.
+As a result, the number of active tokens on the broker is generally less than the number of active sessions (connections).
+ However, keep in mind that this is replicated across all Kafka brokers in the cluster, as clients maintain active sessions to multiple brokers.
```
scholzj, line 705:
```suggestion
However, keep in mind that this is replicated across all Kafka brokers in the cluster, as clients maintain active sessions to multiple brokers.
```

scholzj (approved):
One nit. LGTM otherwise.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/Semaphores.java
```diff
@@ -0,0 +1,33 @@
...
+
+    SemaphoreFuture<T> acquireSemaphore(String key) {
+        return futures.computeIfAbsent(key, v -> new SemaphoreFuture<>());
+    }
+
+    void releaseSemaphore(String key) {
+        futures.remove(key);
+    }
+
+    class SemaphoreFuture<S> extends CompletableFuture<S> {
```
tombentley, line 23: Does this need to extend CF? The API looses some safety because it's possible to get the result of the CF without first acquiring the semaphore.

tombentley (commented):
A minor grip about the api for Semahpores, but otherwise LGTM.

tombentley (approved):

PR #186, opened 2023-03-06
Add support for Java 17

mstruk:
- Added three profiles: java-8, java-11, and java-17 that activate automatically based on the JDK used
- Fixed JavaDoc warnings so the project builds with Java 17
- Fixed `.travis/build.sh` script to use Java 11 and run testsuite on x86_64

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ jobs:
         packages:
         - maven
 jdk:
-- openjdk8
 - openjdk11
 cache:
   directories:
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -18,14 +18,16 @@ exitIfError() {
   [ "$EXIT" != "0" ] && exit $EXIT
 }
 
+arch=$(uname -m)
+
 # The first segment of the version number is '1' for releases < 9; then '9', '10', '11', ...
 JAVA_MAJOR_VERSION=$(java -version 2>&1 | sed -E -n 's/.* version "([0-9]*).*$/\1/p')
 if [ ${JAVA_MAJOR_VERSION} -gt 1 ] ; then
   export JAVA_VERSION=${JAVA_MAJOR_VERSION}
 fi
 
-if [ ${JAVA_MAJOR_VERSION} -eq 1 ] ; then
-  # some parts of the workflow should be done only once on the main build which is currently Java 8
+if [ "$arch" == 'x86_64' ] ; then
+  # some parts of the workflow should be done only once on the main build which is currently Java 11
   export MAIN_BUILD="TRUE"
 fi
 
@@ -40,8 +42,6 @@ else
 fi
 
 mvn spotbugs:check
-
-arch=$(uname -m)
 
 # Also test examples build on different architectures (exclude ppc64le until fixed)
 if [ "$arch" != 'ppc64le' ]; then
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -26,7 +26,7 @@ if [ ${JAVA_MAJOR_VERSION} -gt 1 ] ; then
   export JAVA_VERSION=${JAVA_MAJOR_VERSION}
 fi
 
-if [ "$arch" == 'x86_64' ] ; then
+if [ "$arch" == 'x86_64' ] || [ "$arch" == 'arm64' ] ; then
   # some parts of the workflow should be done only once on the main build which is currently Java 11
   export MAIN_BUILD="TRUE"
 fi
@@ -82,20 +82,24 @@ if [ "${MAIN_BUILD}" == "TRUE" ] ; then
     EXIT=$?
     exitIfError
 
-    clearDockerEnv
-    mvn -e -V -B clean install -f testsuite -Pkafka-3_1_2
-    EXIT=$?
-    exitIfError
+    if [ $JAVA_VERSION -le 11 ]; then
+      clearDockerEnv
+      mvn -e -V -B clean install -f testsuite -Pkafka-3_1_2
+      EXIT=$?
+      exitIfError
 
-    clearDockerEnv
-    mvn -e -V -B clean install -f testsuite -Pkafka-3_0_0
-    EXIT=$?
-    exitIfError
+      clearDockerEnv
+      mvn -e -V -B clean install -f testsuite -Pkafka-3_0_0
+      EXIT=$?
+      exitIfError
 
-    clearDockerEnv
-    mvn -e -V -B clean install -f testsuite -Pkafka-2_8_1
-    EXIT=$?
-    exitIfError
+      clearDockerEnv
+      mvn -e -V -B clean install -f testsuite -Pkafka-2_8_1
+      EXIT=$?
+      exitIfError
+    else
+      echo "Skipped test profiles: kafka-3_1_2, kafka-3_0_0 and kafka-2_8_1"
+    fi
 
     set -e
   fi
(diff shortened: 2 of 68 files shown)


.travis.yml
```diff
@@ -18,7 +18,6 @@ jobs:
         packages:
         - maven
 jdk:
-- openjdk8
 - openjdk11
```
scholzj, line 21: I guess you should add Java 17 here?

.travis/build.sh
```diff
@@ -82,20 +82,24 @@ if [ "${MAIN_BUILD}" == "TRUE" ] ; then
...
-    clearDockerEnv
-    mvn -e -V -B clean install -f testsuite -Pkafka-3_0_0
-    EXIT=$?
-    exitIfError
-
-    clearDockerEnv
-    mvn -e -V -B clean install -f testsuite -Pkafka-2_8_1
-    EXIT=$?
-    exitIfError
+    if [ $JAVA_VERSION -le 11 ]; then
```
scholzj, line 85: Why would you run the tests only on Java 11?

.travis/build.sh
```diff
@@ -18,14 +18,16 @@ exitIfError() {
...
 # The first segment of the version number is '1' for releases < 9; then '9', '10', '11', ...
 JAVA_MAJOR_VERSION=$(java -version 2>&1 | sed -E -n 's/.* version "([0-9]*).*$/\1/p')
 if [ ${JAVA_MAJOR_VERSION} -gt 1 ] ; then
   export JAVA_VERSION=${JAVA_MAJOR_VERSION}
 fi
 
-if [ ${JAVA_MAJOR_VERSION} -eq 1 ] ; then
-  # some parts of the workflow should be done only once on the main build which is currently Java 8
+if [ "$arch" == 'x86_64' ] || [ "$arch" == 'arm64' ] ; then
+  # some parts of the workflow should be done only once on the main build which is currently Java 11
```
scholzj, line 30: Should you check the Java version here?

.travis.yml
mstruk: Yeah, makes sense.

.travis/build.sh
mstruk: Those profiles use Kafka images with Java 11 so they are unable to use the OAuth jar built with Java 17 and latest class version of Java 17. That's why they have to be skipped for Java 17 build.

scholzj: Makes sense, thanks. GitHub didn't show the tests above these lines running on all versions so I did not noticed them.

.travis/build.sh
mstruk: Actually, the question is do we want to run as many of the tests on as many of the platforms as possible.

I'm not sure but IIRC the Arquillian Cube based testsuite only worked on Java 8, that has been fixed by moving to TestContainers.
Another thing I remember is that the testsuite run would take so long or produce so much output that it triggered some limits and was killed by Travis.
I guess we can try again running the full testsuite on both Java 11 and Java 17, and if that will hit limits then I suggest we limit the testsuite to Java 11, that's due to the next paragraph ...

We have to decide which class version libraries we want to push to maven upon release.
I would still prefer Java 8 as that makes the jars most widely usable, but we decided to move on from there. Therefore I suggest Java 11 for the same reason.

scholzj: I don't think anyone is prescribing what Java versions should Oauth support. We need Java 11 as a minimum. But if you want to support Java 8, then fine. But Java 8 is super old, Kafka will stop supporting it soon etc. So I think it is time to cut it. But even if you want to release with Java 8 as a target version, we need to have testing with Java 11 and Java 17

mstruk: Ok, I added `openjdk8` back to the Travis build, and reworked logic so that testsuite is run for all java versions on all platforms. Only Java 8 build on `x86_64` will be published if it's a release build.

.travis/build.sh
```diff
@@ -18,70 +18,58 @@ exitIfError() {
   [ "$EXIT" != "0" ] && exit $EXIT
 }
 
+arch=$(uname -m)
+echo "arch: $arch"
```
scholzj, line 22: If you want to keep this, maybe it can be a bit nicer? `Architecture: `?

pom.xml
```diff
@@ -344,6 +343,38 @@
...
+            <id>java-17</id>
+            <activation>
+                <jdk>[17,</jdk>
+            </activation>
+            <properties>
+                <maven.compiler.source>11</maven.compiler.source>
+                <maven.compiler.target>17</maven.compiler.target>
+                <maven.compiler.release>17</maven.compiler.release>
+            </properties>
+        </profile>
```
scholzj, line 377: Both here and in the other pom.xml ... Do we really need the profiles here? The main point of these Java 11 and Java 17 builds is to ensure that it works in the environments using Java 11 and Java 17. So I do not think we need to change the source / target versions as we want it to be as close as possible to the code you will actually release.

I do not think we did anything like this on the other Strimzi projects either.

mstruk: I found that `maven.compiler.release` option was necessary to build with Java 11 or 17. Without it I was getting an error like: `system modules path not set in conjunction with -source`

However, Java-8 build would not build if this option was present.

It could be the specifics of my local environment, but I just did whatever was necessary to get things to work.

The other thing there is the compiler source 11 in java-17 profile.  This one I kind of forgot in there.
I'll set it to 17 so it's in sync with the others.

.travis/build.sh
mstruk: Yes, let's have the printout to be sure that the actual architecture label is not something different then I'm expecting.
I can make it a bit nicer :)

pom.xml
scholzj: Not sure what `maven.compiler.release` does TBH. But maybe we can at least keep the source and target?

scholzj: I mean ideally, we would want to have a single build and then test the binaries it produces with the different JDKs. But that would be quite complicated. But I think it would be great to get at least close to it.

mstruk: In that case I would most prefer`1.8`. We could remove `java-8` profile, and remove source and target overrides from `java-11` and `java-18` profiles, but they would still need the release set to `8`.
Would that be better?

mstruk:
I modified the profiles and versions in `pom.xml` as described above, and fixed hostname aliases as `.travis/build.sh` so that a full testsuite run is passing on all java versions on `amd64`, and on the configured Java 11 version on `s390x`, where as the testsuite run is skipped on `ppc64le` due to unavailable docker images for that platform.

testsuite/pom.xml
```diff
@@ -383,5 +399,38 @@
...
+            <id>java-17</id>
+            <activation>
+                <jdk>[17,</jdk>
+            </activation>
+            <properties>
+                <maven.compiler.source>11</maven.compiler.source>
+                <maven.compiler.target>17</maven.compiler.target>
+                <maven.compiler.release>17</maven.compiler.release>
+            </properties>
+        </profile>
```
scholzj, line 433: Should this use the same profile as the main `pom.xml`? Or is there some reason to have it different?

pom.xml
```diff
@@ -344,6 +343,24 @@
...
+        <profile>
+            <id>java-17</id>
+            <activation>
+                <jdk>[17,</jdk>
+            </activation>
+            <properties>
+                <maven.compiler.release>8</maven.compiler.release>
+            </properties>
+        </profile>
```
scholzj, line 363: I do not get why is this needed if at the end the settings are anyway the same. Can't you just keep `<maven.compiler.release>8</maven.compiler.release>` in the main properties?

scholzj (approved):
Left two nits / questions. But if you think it is needed as it s, then its fine with me.

pom.xml
scholzj: Ok, thanks,

testsuite/pom.xml
mstruk: Ugh, missed that. Good catch.

PR #187, opened 2023-03-07
Update examples

mstruk:
- Fix missing IdempotentWrite and UI instructions in Kafka authorization example
- Bump versions of Kafka image used


testsuite/docker/kafka/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/strimzi/kafka:0.31.1-kafka-3.2.3
+FROM quay.io/strimzi/kafka:0.33.1-kafka-3.3.2
```
scholzj, line 1: In case you want to use the latest ...

```suggestion
FROM quay.io/strimzi/kafka:0.33.2-kafka-3.4.0
```

scholzj (commented):
The latest versions are Strimzi 0.33.2 and Kafka 3.4.0 ... would it make sense to use those instead of 0.33.1 and 3.3.2?

testsuite/docker/kafka/Dockerfile
mstruk: I bumped this all around the examples and the testsuite, added a new profile and added it to Travis build.

scholzj (approved):

PR #188, opened 2023-03-07
Fix KeycloakRBACAuthorizer to work with `StandardAuthorizer` in KRAFT mode

mstruk:


mstruk:
I've been having a difficulty getting the testsuite to work in KRAFT mode. The broker starts fine and I can see in the log that KRAFT mode is detected and the `StandardAuthorizer` is used as the delegate. The authorization requests are working as they are supposed to based on the log.

But for some reason only the controller listener port is responsive, the reason that there are some requests at all as the broker connects to itself and all those connections make some requests that go through the authorizer. 
The broker listeners ports are accepting connections but requests just hang on them. I'm trying to find the reason.

In any case this PR even though without the test can only improve the behaviour in KRAFT mode, since without it the KeycloakRBACAuthorizer simply crashes with exception in KRAFT mode.

scholzj:
@mstruk How do you run the KRaft cluster? I have a simple Docker Compose file for KRaft based on the Strimzi image here: https://github.com/scholzj/strimzi-compose-up/blob/main/kraft/docker-compose.yaml if that helps

mstruk:
I do all these steps in my own script: https://github.com/mstruk/strimzi-kafka-oauth/blob/kraft-tests/testsuite/docker/kafka/scripts/start-kraft.sh

But then I have a lot more complex configuration in terms of listeners: https://github.com/mstruk/strimzi-kafka-oauth/blob/kraft-tests/testsuite/keycloak-authz-kraft-tests/docker-compose.yml

There must be some little detail somewhere in there, but it's frustrating how it is even possible to have no exception at all. I'll have to start from complete scratch and see it work, and then add things one at a time until it breaks.

mstruk:
That will take a day though, so I'm for merging this without a testsuite for now. As I said, it can't make anything worse than it already is in KRAFT mode.

mstruk:
Since the tests have passed it means the 'zookeeper' mode is still working as it's supposed to.

scholzj (commented):
I think this is done a bit in a hurry ... I think we need to figure out for example:
* What it does during an upgrade
* Figure out how to deal with early-start listeners

mstruk:
Very well. This won't make it into 0.12.0 then.

I'll need to better understand the two issues you are mentioning - upgrade, and early-start listeners.

mstruk:
I opend this PR for a review. I'm not sure the following two points should be a concern of this PR:

> * What it does during an upgrade
> * Figure out how to deal with early-start listeners

Authorizer is instantiated and configured by Kafka broker and lifecycle methods like start(), close(), configure() are called on it. Authorizer is instantiated independently of the listeners and after the listeners are configured. I guess if the listeners are updated at runtime the broker decides to what extent that impacts the authorizer and some lifecycle methods are invoked in a certain order or nothing is done at all. Currently my assumption is that things will "just work". If it turns out that is not the case we can address it in another PR.

scholzj:
> Authorizer is instantiated and configured by Kafka broker and lifecycle methods like start(), close(), configure() are called on it. Authorizer is instantiated independently of the listeners and after the listeners are configured. I guess if the listeners are updated at runtime the broker decides to what extent that impacts the authorizer and some lifecycle methods are invoked in a certain order or nothing is done at all. Currently my assumption is that things will "just work". If it turns out that is not the case we can address it in another PR.

I think you have to test it. Leaving it to another PR is fine, but I would expect any other tests to be done only after the release and not after the PR unless you do them.

scholzj:
Or alternatively, it might not work in KRaft in that release, which is something we can live with as well if needed I guess as it is anyway not production ready 🤷

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/GrantsHandler.java
```diff
@@ -0,0 +1,520 @@
...
+            }
+
+            for (GrantsHandler.Future f : scheduled) {
+                try {
+                    f.get();
+                } catch (ExecutionException e) {
+                    final Throwable cause = e.getCause();
+                    if (cause instanceof HttpException) {
+                        log.debug("[IGNORED] Failed to fetch grants for user: {}", cause.getMessage());
+                        if (401 == ((HttpException) cause).getStatus()) {
```
tombentley, line 391: Comment that 401==Unauthorized

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/GrantsHandler.java
```diff
@@ -0,0 +1,520 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Function;
+
+import static io.strimzi.kafka.oauth.common.JSONUtil.asSetOfNodes;
+import static io.strimzi.kafka.oauth.common.LogUtil.mask;
+
+/**
+ * The class that handles grants cache and services to maintain it
```
tombentley, line 38: I think somewhere, and it may as well be here in the Javadoc, you need to explain clearly how the threading and locking in this class works. You have `synchronized` access on the grants cache, you have three different `Executors`, you have a `Semaphore` abstraction.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/GrantsHandler.java
```diff
@@ -0,0 +1,520 @@
...
+     */
+    private JsonNode fetchGrantsWithRetry(String token) {
+
+        int i = 0;
+        do {
+            i += 1;
+
+            try {
+                if (i > 1) {
+                    log.debug("Grants request attempt no. " + i);
```
tombentley, line 310: Use `{}` placeholder.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/GrantsHandler.java
```diff
@@ -0,0 +1,520 @@
...
+            } catch (Exception e) {
+                if (e instanceof HttpException) {
+                    int status = ((HttpException) e).getStatus();
+                    if (403 == status || 401 == status) {
+                        throw e;
+                    }
+                }
+
+                if (log.isInfoEnabled()) {
+                    log.info("Failed to fetch grants on try no. " + i, e);
```
tombentley, line 323: Use `{}` placeholder.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/GrantsHandler.java
```diff
@@ -0,0 +1,520 @@
...
+                String userId = ent.getKey();
+                Info grantsInfo = ent.getValue();
+                if (grantsInfo.getLastUsed() < System.currentTimeMillis() - grantsMaxIdleMillis) {
+                    log.debug("Skipping refreshing grants for user '{}' due to max idle time.", userId);
+                    removeUserFromCacheIfExpiredOrIdle(userId);
+                }
+                scheduled.add(new Future(userId, grantsInfo, refreshWorker.submit(() -> {
+
+                    if (log.isTraceEnabled()) {
+                        log.trace("Fetch grants for user: " + userId + ", token: " + mask(grantsInfo.getAccessToken()));
```
tombentley, line 357: Use `{}` placeholder.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/GrantsHandler.java
```diff
@@ -0,0 +1,520 @@
...
+                        log.debug("[IGNORED] Failed to fetch grants for user: {}", cause.getMessage());
+                        if (401 == ((HttpException) cause).getStatus()) {
+                            grantsCache.remove(f.getUserId());
+                            log.debug("Removed user from grants cache: {}", f.getUserId());
+                            Services.getInstance().getSessions().removeAllWithMatchingAccessToken(f.getGrantsInfo().accessToken);
+                            continue;
+                        }
+                    }
+                    if (log.isWarnEnabled()) {
+                        log.warn("[IGNORED] Failed to fetch grants for user: " + e.getMessage(), e);
```
tombentley, line 399: Use `{}` placeholder.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/GrantsHandler.java
```diff
@@ -0,0 +1,520 @@
...
+                    if (cause instanceof HttpException) {
+                        log.debug("[IGNORED] Failed to fetch grants for user: {}", cause.getMessage());
+                        if (401 == ((HttpException) cause).getStatus()) {
+                            grantsCache.remove(f.getUserId());
+                            log.debug("Removed user from grants cache: {}", f.getUserId());
+                            Services.getInstance().getSessions().removeAllWithMatchingAccessToken(f.getGrantsInfo().accessToken);
+                            continue;
+                        }
+                    }
+                    if (log.isWarnEnabled()) {
```
tombentley, line 398: Is the call to `e.getMessage()` so expensive that this guard is worth the noise?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/GrantsHandler.java
```diff
@@ -0,0 +1,520 @@
...
+                        log.warn("[IGNORED] Failed to fetch grants for user: " + e.getMessage(), e);
+                    }
+
+                } catch (Throwable e) {
+                    if (log.isWarnEnabled()) {
+                        log.warn("[IGNORED] Failed to fetch grants for user: " + f.getUserId() + ", token: " + mask(f.getGrantsInfo().accessToken) + " - " + e.getMessage(), e);
```
tombentley, line 404: Use `{}` placeholder.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/GrantsHandler.java
```diff
@@ -0,0 +1,520 @@
...
+                    }
+                }
+            }
+
+        } catch (Throwable t) {
+            // Log, but don't rethrow the exception to prevent scheduler cancelling the scheduled job.
+            log.error(t.getMessage(), t);
```
tombentley, line 411: If `t.getMessage()` contained `{}` then message that was logged would be different than if `t.getMessage()` didn't contain `{}`. So this should probably be `log.error("{}", t.getMessage(), t);` ?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java
```diff
@@ -0,0 +1,222 @@
...
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.atomic.AtomicInteger;
+
+
+/**
+ * An authorizer that ensures a single stateful instance is used for authorization callbacks.
```
tombentley, line 35: I read this and wondered "single stateful instance _of what_?". And surely there's more to this class than the single stateful instance part: What does that instance _do_?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java
```diff
@@ -0,0 +1,222 @@
...
+ */
+public class KeycloakAuthorizer implements ClusterMetadataAuthorizer {
+
+    private static final Logger log = LoggerFactory.getLogger(KeycloakAuthorizer.class);
+
+    private final static AtomicInteger VERSION_COUNTER = new AtomicInteger(1);
+    private final int version = VERSION_COUNTER.getAndIncrement();
```
tombentley, line 42: Wouldn't this be clearer:

```suggestion
    private final static AtomicInteger VERSION_COUNTER = new AtomicInteger(2);
    private final int version = 1;
```

scholzj (commented):
I left some commnets. It would be also good to have a proper description of the PR summarizing the changes it contains. As piecing it together from the README changes and the code is quite confusing

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/GrantsHandler.java
mstruk: Whoa, you just blew my mind here! You are absolutely right! I never understood this properly, and it went by me so easy.
There should be a Spotbugs rule for this.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java
mstruk: Ok, we need more explanation here how this class works and what is its relationship to KeycloakRBACAuthorizer.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java
mstruk: The intention of this code is for created instances to be versioned. When in KRaft mode multiple instances of the authorizer can be instantiated. 

Your suggestion would assign them all version == 1, and VERSION_COUNTER is then redundant and it's never used. 

My intention is for each new instance to get the next incremental number. You can then easily track the instances via toString() and also you have some idea how many there have been instantiated. This `version` is not serialization version.

README.md
```diff
@@ -100,7 +100,7 @@ Authorization in Kafka is implemented completely separately and independently of
 Thus, it is possible to configure Kafka brokers to use OAuth2 based authentication, and at the same time the default ACL authorization. 
 
 Since version 0.3.0 Strimzi Kafka OAuth provides token-based authorization using Keycloak as authorization server, and taking advantage of [Keycloak Authorization Services](https://www.keycloak.org/docs/latest/authorization_services/).
-See [examples authorization README](examples/README-authz.md) for a demonstration on how to install, and use [KeycloakRBACAuthorizer](oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java) which implements this functionality.
+See [examples authorization README](examples/README-authz.md) for a demonstration on how to install, and use [KeycloakAuthorizer](oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java) which implements this functionality.
```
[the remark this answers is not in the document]
scholzj: Ok, that makes sense, thanks for the explanation. But I think it is really important to explain it in the release notes and in the PR description ;-).

scholzj (approved):

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/Configuration.java
```diff
@@ -0,0 +1,470 @@
...
+     */
+    Configuration(Map<String, ?> configs) {
+
+        this.configMap = configs;
+
+        AuthzConfig authzConfig = convertToAuthzConfig(configs);
+
+        String pbclass = (String) configMap.get("principal.builder.class");
+        if (!PRINCIPAL_BUILDER_CLASS.equals(pbclass) && !DEPRECATED_PRINCIPAL_BUILDER_CLASS.equals(pbclass)) {
+            throw new ConfigException("KeycloakRBACAuthorizer requires " + PRINCIPAL_BUILDER_CLASS + " as 'principal.builder.class'");
```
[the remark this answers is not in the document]
mstruk: Opted for avoiding the authorizer name altogether.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java
mstruk: I renamed the variables to minimise confusion.

mstruk:
I hope I sufficiently addressed the comments and proposals.
I'm still looking at Travis CI failures if it's an intermittent failure or something specific to Java 8.

Other than that, is there something else that needs fixing or getting addressed better within this PR or separately?

tombentley:
@mstruk I'll try to take a look by this time tomorrow.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
```diff
@@ -160,258 +156,123 @@
...
+        configuration = new Configuration(configs);
+        configuration.printLogs();
 
-        authzSensorKeyProducer = new KeycloakAuthorizationSensorKeyProducer("keycloak-authorizer", tokenEndpointUrl);
-        grantsSensorKeyProducer = new GrantsHttpSensorKeyProducer("keycloak-authorizer", tokenEndpointUrl);
-
-        if (delegateToKafkaACL) {
-            super.configure(configs);
-        }
+        instantiateObjects(configuration);
```
tombentley, line 198: It's really `assignFields`, and I think that method can be `private`.

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java
```diff
@@ -160,258 +156,123 @@
  * This authorizer honors the <em>super.users</em> configuration. Super users are automatically granted any authorization request.
  * </p>
  */
-@SuppressWarnings("deprecation")
-public class KeycloakRBACAuthorizer extends AclAuthorizer {
-
-    private static final String PRINCIPAL_BUILDER_CLASS = OAuthKafkaPrincipalBuilder.class.getName();
-    private static final String DEPRECATED_PRINCIPAL_BUILDER_CLASS = JwtKafkaPrincipalBuilder.class.getName();
+public class KeycloakRBACAuthorizer implements Authorizer {
```
tombentley, line 159: I guess we can't deprecate this class because the `KeycloakAuthorizer` can delegate to it. But how are we expecting a user to know they need to move to `KeycloakAuthorizer`?

tombentley (approved):

PR #190, opened 2023-04-03
Bump jsonpath dependency to 2.8.0 for sub-dependency CVE

ajborley:
Mitigates CVE-2023-1370, which is found in json-smart, which is a sub-dependency 
of the com.jayway.jsonpath:json-path dependency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@
         <kafka.version>3.3.1</kafka.version>
         <jackson.version>2.13.4</jackson.version>
         <jackson.databind.version>2.13.4.2</jackson.databind.version>
-        <jsonpath.version>2.6.0</jsonpath.version>
+        <jsonpath.version>2.8.0</jsonpath.version>
         <junit.version>4.13.1</junit.version>
         <slf4j.version>1.7.36</slf4j.version>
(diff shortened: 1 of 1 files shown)


scholzj (approved):
LGTM assuming the tests pass.

scholzj:
It looks good to me. But I think @mstruk mentioned some issues with backwards compatibility between 2.6.0 and 2.8.0 (if I'm not mistaken and it was this dependency).

mstruk:
The incompatibility is actually the reason the test is failing. They fixed some bug in how mid-evaluation exception was used for flow control and resulted in non-intuitive / buggy results when testing attribute value for null using `==`, `!=`. The problem was that when the attribute was not present then `== null` would result in `false`, and `!= null` would result in `true`.
For this reason we have in the [README doc](https://github.com/strimzi/strimzi-kafka-oauth/#custom-claim-checking) a warning about not using `null` in custom claim check queries. But people don't always read the doc ...

But then, we can't be stuck on a CVE-ed version. I'm thinking of checking for the presence of `null` in the `oauth.custom.claim.check` query, and printing a warning during startup. I guess there's not much more we can do.

mstruk:
This issue is addressed by #196.

PR #193, opened 2023-06-28
Remove the use of `metric.reporters` in OAuth metrics, use `strimzi.metric.reporters` instead

mstruk:
This PR removes the use of `metric.reporters` in OAuth metrics to configure the reporters to instantiate.
The reason is that OAuth metrics integrates with the reporters by creating new instances of them, resulting in one instance being created by Kafka broker, and another instance created by OAuth metrics. Not all reporters work properly when multiple instances are instantiated.

With this PR OAuth metrics, when enabled, will use `strimzi.metric.reporters` as a configuration of which reporters to instantiate.

Also, previously `org.apache.kafka.common.metrics.JmxReporter` was automatically instantiated every time, mirroring how Kafka handles `metric.reporters`. With Kafka 4.0 the behaviour of `metric.reporters` is set to change and only instantiate explicitly listed reporters. `strimzi.metric.reporters` already implements this behavour for OAuth metrics regardless of the Kafka version used.

In order to migrate existing broker configuration add the following env variable:
```
STRIMZI_METRIC_REPORTERS=org.apache.kafka.common.metrics.JmxReporter
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1225,21 +1225,38 @@ Configuring the metrics
 
 By default, the gathering and exporting of metrics is disabled. Metrics are available to get an insight into the performance and failures during token validation, authorization operations and client authentication to the authorization server. You can also monitor the authorization server requests by background services such as refreshing of JWKS keys and refreshing of grants when `KeycloakAuthorizer` is used.
 
-You can enable metrics for token validation on the Kafka broker or for client authentication on the client by setting the following JAAS option to `true`:
+You can enable metrics for token validation, and keycloak authorization on the Kafka broker or for client authentication on the client by setting the following JAAS option to `true`:
 - `oauth.enable.metrics` (e.g.: "true")
 
-You can enable metrics for `KeycloakAuthorizer` by setting an analogous option in Kafka broker's `server.properties` file:
+You can also enable metrics only for keycloak authorization by setting an analogous option in Kafka broker's `server.properties` file:
 - `strimzi.authorization.enable.metrics` (e.g.: "true")
 
-If `OAUTH_ENABLE_METRICS` env variable is set or if `oauth.enable.metrics` system property is set, that will both also enable the metrics for `KeycloakAuthorizer`.
+If `OAUTH_ENABLE_METRICS` env variable is set or if `oauth.enable.metrics` system property is set, that will also enable the metrics for keycloak authorization (as well as for token validation, and client authentication).
 
+The OAuth metrics ignores the Kafka `metric.reporters` option in order to prevent automatically instantiating double instances of reporters. Most reporters may expect that they are singleton object and may not function properly in multiple copies.
+Instead, there is `strimzi.metric.reporters` option where the reporters that support multiple copies can be specified for the purpose of metrics integration:
+- `strimzi.metric.reporters` (e.g.: "org.apache.kafka.common.metrics.JmxReporter", use ',' as a separator to enable multiple reporters)
+
+The configuration option `strimzi.metric.reporters` has to be configured as an env variable or a system property. Using it on the Kafka broker inside `server.properties` does not work reliably due to multiple pluggability mechanisms that can be used (authorizer, authentication callback handler, client).
+Some of these mechanisms get the `server.properties` filtered so only configuration recognised by Kafka makes it through. However, this is a global OAuth Metrics configuration, and it is initialized on first use by any of the components, using the configuration provided to that component.
+Specifically, the inter-broker client using OAUTHBEARER might be the first to trigger OAuth Metrics initialisation on the broker, and does not see this config option.
+
+In order to reliably configure `strimzi.metric.reporters` one of the following options should be used when starting a Kafka broker:
+- `STRIMZI_METRIC_REPORTERS` env variable
+- `strimzi.metric.reporters` env variable or system property
+
+When OAuth metrics are enabled the OAuth layer can not reuse existing metric reporter instances, and have to create its own copies of metric reporters. Noteably, it does not create its copy of `org.apache.kafka.common.metrics.JmxReporter` automatically. Instead, it has to be explicitly specified in `strimzi.metric.reporters` configuration as item in the list of reporters to instantiate.
+At the moment there is no way to integrate with the existing Kafka metrics / reporters objects already instantiated in different Kafka runtimes (producer, consumer, broker, ...).
+
+NOTE: This breaks compatibility with OAuth versions preceding 0.13.0 where `metric.reporters` was used to configure reporters which were consequently automatically instantiated twice.
+
+
+The Kafka options that control sampling are honored: `metrics.num.samples`, `metrics.recording.level` and `metrics.sample.window.ms`.  
+
+A common use-case is for metrics to be exposed through JMX managed beans. They can then also be exposed as Prometheus metrics by using the Prometheus JMX Exporter agent, and mapping the JMX metrics names to prometheus metrics names.
 If `oauth.config.id` is specified in JAAS configuration of the listener or the client, it will be available in MBean / metric name as `contextId` attribute. If not specified, it will be calculated from JAAS configuration for the validator or default to `client` in client JAAS config, or `keycloak-authorizer` for KeycloakAuthorizer metrics.
 
-Metrics are exposed through JMX managed beans. They can also be exposed as Prometheus metrics by using the Prometheus JMX Exporter agent, and mapping the JMX metrics names to prometheus metrics names.
-
-The OAuth metrics also honor the `metric.reporters`, `metrics.num.samples`, `metrics.recording.level` and `metrics.sample.window.ms` configurations of Kafka runtimes. When OAuth metrics are enabled the OAuth layer creates duplicate instances of `JmxReporter` and the configured `MetricReporter`s since at the moment there is no other way to integrate with the existing metrics system of Kafka runtimes. 
-
-When OAuth metrics are enabled, managed beans are registered on demand, containing the attributes that are easily translated into Prometheus metrics.
+When `JmxReporter` is enabled, managed beans are registered on demand, containing the attributes that are easily translated into Prometheus metrics.
 
 Each registered MBean contains two counter variables - `count`, and `totalTimeMs`.
 It also contains three gauge variables - `minTimeMs`, `maxTimeMs` and `avgTimeMs`. These are measured within the configured sample time window.
--- a/oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
+++ b/oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
@@ -10,7 +10,6 @@ import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
-import org.apache.kafka.common.metrics.JmxReporter;
 import org.apache.kafka.common.metrics.KafkaMetricsContext;
 import org.apache.kafka.common.metrics.MetricConfig;
 import org.apache.kafka.common.metrics.Metrics;
@@ -24,13 +23,17 @@ import org.apache.kafka.common.metrics.stats.Min;
 import org.apache.kafka.common.utils.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
+import static io.strimzi.kafka.oauth.metrics.GlobalConfig.STRIMZI_METRIC_REPORTERS;
 import static org.apache.kafka.clients.CommonClientConfigs.CLIENT_ID_CONFIG;
 import static org.apache.kafka.clients.CommonClientConfigs.METRICS_NUM_SAMPLES_CONFIG;
 import static org.apache.kafka.clients.CommonClientConfigs.METRICS_RECORDING_LEVEL_CONFIG;
@@ -38,9 +41,22 @@ import static org.apache.kafka.clients.CommonClientConfigs.METRICS_SAMPLE_WINDOW
 
 /**
  * The singleton for handling a cache of all the Sensors to prevent unnecessary redundant re-registrations.
- * There is a one-to-one mapping between a SensorKey and a Sensor, and one-to-one mapping between a Sensor and an MBean name.
- *
- * MBeans are registered as requested by JmxReporter attached to the Metrics object.
+ * There is a one-to-one mapping between a <code>SensorKey</code> and a <code>Sensor</code>, and one-to-one mapping between a <code>Sensor</code> and an <code>MBean</code> name.
+ * <p>
+ * MBeans are registered as requested by <code>JmxReporter</code> attached to the <code>Metrics</code> object.
+ * The <code>JmxReporter</code> has to be explicitly configured using a config option <code>strimzi.metric.reporters</code>, which is
+ * analogous to the Kafka <code>metric.reporters</code> configuration option. It is not configured by default.
+ * <p>
+ * Since OAuth instantiates its own <code>Metrics</code> object it also has to instantiate reporters to attach them to this <code>Metrics</code> object.
+ * To prevent double instantiation of <code>MetricReporter</code> objects that require to be singleton, all <code>MetricReporter</code> objects
+ * to be integrated with <code>OAuthMetrics</code> have to be separately instantiated.
+ * <p>
+ * Example:
+ * <pre>
+ *    strimzi.metric.reporters=org.apache.kafka.common.metrics.JmxReporter
+ * </pre>
+ * Note: On the Kafka broker it is best to use <code>STRIMZI_METRIC_REPORTERS</code> env variable or <code>strimzi.metric.reporters</code> system property,
+ * rather than a `server.properties` global configuration option.
  */
 public class OAuthMetrics {
 
@@ -57,9 +73,14 @@ public class OAuthMetrics {
      *
      * @param configMap Configuration properties
      */
+    @SuppressWarnings("unchecked")
     OAuthMetrics(Map<String, ?> configMap) {
-        this.configMap = configMap;
         this.config = new Config(configMap);
+
+        // Make sure to add the resolved 'strimzi.metric.reporters' configuration to the config map
+        ((Map<String, Object>) configMap).put(STRIMZI_METRIC_REPORTERS, config.getValue(STRIMZI_METRIC_REPORTERS));
+        this.configMap = configMap;
+
         this.metrics = initKafkaMetrics();
     }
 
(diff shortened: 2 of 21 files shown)


scholzj (commented):
Where will this new option `strimzi.metrics.reporter` be used? If I got it rght and it is in the general configuration of Kafka, should it rather be a `strimzi.oauth.metrics.reporter`?

scholzj:
I do not think the `strimzi.` itself is an issue. Its more without the `.oauth` behind it, it possibly collides with other Strimzi options. So if it is something only Oauth reads as far as I understood, I think `strimzi.oauth...` would be better if we stick with the `strimzi.` part. Not sure if @tombentley for example has some thoughts about it.

tombentley:
`strimzi.oauth...` seems better to me.

testsuite/common/src/main/java/io/strimzi/testsuite/oauth/common/TestUtil.java
```diff
@@ -93,4 +93,32 @@ public static void logStart(String msg) {
         System.out.println("========    "  + msg);
         System.out.println();
     }
+
+    public static int findFirstMatchingInLog(List<String> log, String regex) {
+        int lineNum = 0;
+        Pattern pattern = Pattern.compile(prepareRegex(regex));
+        for (String line: log) {
+            if (pattern.matcher(line).matches()) {
```
tombentley, line 101: It seems a bit weird to have `prepareRegex()` and then use `Matcher.matches()`, rather than just using `Matcher.find()`. Using `find()` would be the preferred way to do this, but if there's a good reason for it then I think it at least deserves a comment.

README.md
```diff
@@ -1225,21 +1225,38 @@ Configuring the metrics
...
+In order to reliably configure `strimzi.metric.reporters` one of the following options should be used when starting a Kafka broker:
+- `STRIMZI_METRIC_REPORTERS` env variable
+- `strimzi.metric.reporters` env variable or system property
+
+When OAuth metrics are enabled the OAuth layer can not reuse existing metric reporter instances, and have to create its own copies of metric reporters. Noteably, it does not create its copy of `org.apache.kafka.common.metrics.JmxReporter` automatically. Instead, it has to be explicitly specified in `strimzi.metric.reporters` configuration as item in the list of reporters to instantiate.
+At the moment there is no way to integrate with the existing Kafka metrics / reporters objects already instantiated in different Kafka runtimes (producer, consumer, broker, ...).
 
-Metrics are exposed through JMX managed beans. They can also be exposed as Prometheus metrics by using the Prometheus JMX Exporter agent, and mapping the JMX metrics names to prometheus metrics names.
+NOTE: This breaks compatibility with OAuth versions preceding 0.13.0 where `metric.reporters` was used to configure reporters which were consequently automatically instantiated twice.
```
tombentley, line 1251: Is this compatibility break _unavoidable_?

README.md
```diff
@@ -1225,21 +1225,38 @@ Configuring the metrics
...
+- `strimzi.metric.reporters` (e.g.: "org.apache.kafka.common.metrics.JmxReporter", use ',' as a separator to enable multiple reporters)
+
+The configuration option `strimzi.metric.reporters` has to be configured as an env variable or a system property. Using it on the Kafka broker inside `server.properties` does not work reliably due to multiple pluggability mechanisms that can be used (authorizer, authentication callback handler, client).
+Some of these mechanisms get the `server.properties` filtered so only configuration recognised by Kafka makes it through. However, this is a global OAuth Metrics configuration, and it is initialized on first use by any of the components, using the configuration provided to that component.
+Specifically, the inter-broker client using OAUTHBEARER might be the first to trigger OAuth Metrics initialisation on the broker, and does not see this config option.
+
+In order to reliably configure `strimzi.metric.reporters` one of the following options should be used when starting a Kafka broker:
+- `STRIMZI_METRIC_REPORTERS` env variable
+- `strimzi.metric.reporters` env variable or system property
+
```
tombentley, line 1247: This is a lot of verbose description which isn't completely easy to understand, but which people would need to follow in order to make use of this feature. I don't see anywhere actual examples they they could copy and paste and expect to work. Could we add some examples somewhere?

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
```diff
@@ -90,26 +111,48 @@ private Metrics initKafkaMetrics() {
...
     private List<MetricsReporter> initReporters() {
         AbstractConfig kafkaConfig = initKafkaConfig();
-        List<MetricsReporter> reporters = kafkaConfig.getConfiguredInstances(CommonClientConfigs.METRIC_REPORTER_CLASSES_CONFIG,
-                MetricsReporter.class);
-
-        JmxReporter reporter = new JmxReporter();
-        reporter.configure(configMap);
-
-        reporters.add(reporter);
+        List<MetricsReporter> reporters = new ArrayList<>(3);
```
mimaison, line 114: Why set the starting size to 3?

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
```diff
@@ -90,26 +111,48 @@ private Metrics initKafkaMetrics() {
...
+        if (configMap.get(STRIMZI_METRIC_REPORTERS) != null) {
+            reporters = kafkaConfig.getConfiguredInstances(STRIMZI_METRIC_REPORTERS, MetricsReporter.class);
+        } else {
+            log.warn("Configuration option '{}' is not set, OAuth metrics will not be exported to JMX", STRIMZI_METRIC_REPORTERS);
```
mimaison, line 118: We define the config as low importance below and without a default value but if it's not set we emit a warning. Should we increase the importance or downgrade that log level?

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
```diff
@@ -90,26 +111,48 @@ private Metrics initKafkaMetrics() {
...
+        for (Map.Entry<String, ?> ent: configMap.entrySet()) {
+            Object val = ent.getValue();
+            if (val == null) {
+                continue;
+            }
+            if (val instanceof Class) {
+                result.put(ent.getKey(), ((Class<?>) val).getCanonicalName());
+            } else if (val instanceof List) {
+                String stringVal = ((List<?>) val).stream().map(String::valueOf).collect(Collectors.joining(","));
+                if (!"".equals(stringVal)) {
```
mimaison, line 148: Can we use `isEmpty()`?

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
```diff
@@ -90,26 +111,48 @@ private Metrics initKafkaMetrics() {
...
+
+    private Map<String, String> toMapOfStringValues(Map<String, ?> configMap) {
+        HashMap<String, String> result = new HashMap<>();
+        for (Map.Entry<String, ?> ent: configMap.entrySet()) {
+            Object val = ent.getValue();
+            if (val == null) {
+                continue;
+            }
+            if (val instanceof Class) {
+                result.put(ent.getKey(), ((Class<?>) val).getCanonicalName());
```
mimaison, line 145: Could we use the new [instanceof pattern matching](https://openjdk.org/jeps/394) syntax here?

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/Services.java
```diff
@@ -39,6 +39,13 @@ public static synchronized void configure(Map<String, ?> configs) {
         }
     }
 
+    /**
+     * Close any configured Services so they can be reinitialised again
+     */
+    public static synchronized void close() {
+        services = null;
```
mimaison, line 46: This is not really _closing_ services

README.md
mstruk: Not breaking compatibility means to keep checking if 'metric.reporters' is set and if it is set instantiating the configured reporters. That is what happens when we configure CruiseControl metric reporter which results in exceptions at start time. We could say 'if `strimzi.oauth.metric.reporters` is NOT set, behave as before - if `metric.reporters` was set instantiate the reporters mentioned there, if `JmxReporter` is not mentioned instantiate it anyway, except if running on Kafka 3.5.0 in which case check if `auto.include.jmx.reporter` is set to false in which case don't instantiate it (but in the future with Kafka 4.0 this check will have to be modified again as `auto.include.jmx.reporter` will no longer be taken into account, and `JmxReporter` will not be automatically instantiated if not mentioned in `metric.reporters`.
The result of us maintaining compatibility is that we don't address automatic double instantiation and people may face exceptions and startup issues as a result of using `oauth.enable.metrics` or `strimzi.authorization.enable.metrics`, like we face it in Strimzi when CruiseControl is enabled.
By breaking this backwards compatibility we can ensure that no automatic double instantiation occurs as user has to not only enable oauth metrics, but also specify the reporters to instantiate. Also we can immediately move on to use Kafka 4.0 configuration model for 'out' metric reporters which is where Kafka is moving.
If we keep providing backwards support the question is then do we also ensure that in the future Kafka 4.0 the behaviour will mimic that of Kafka.
I clearly lean towards breaking backwards support but if it's deemed a 'no-no' in that we'll do whatever it takes to not break backwards compatibility then it is possible to keep it.

README.md
mstruk: There are different ways to configure it.

The simplest one would be:
```
# Enable OAuth metrics for all listeners, Keycloak authorizer, and inter-broker clients:
export OAUTH_ENABLE_METRICS=true
export STRIMZI_OAUTH_METRIC_REPORTERS=org.apache.kafka.common.metrics.JmxReporter
bin/kafka-server-start.sh config/server.properties
```

I wouldn't go more complex with it. Let people read the docs. If the docs are not clear enough we'll improve.

You can have more complex config for different contexts. For example for the Kafka client you can have:
```
~/client.properties:
...
sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
  oauth.token.endpoint.uri="https://server/token-endpoint" oauth.client.id="clientId" oauth.client.secret="client-secret" oauth.enable.metrics="true";

strimzi.oauth.metric.reporters=org.apache.kafka.common.metrics.JmxReporter
...
```

Start client:
```
bin/kafka-console-producer.sh --broker-list kafka:9092 --topic my-topic  --producer.config=$HOME/client.properties
```

But it could also simply be:
```
export OAUTH_ENABLE_METRICS=true
export STRIMZI_OAUTH_METRIC_REPORTERS=org.apache.kafka.common.metrics.JmxReporter
bin/kafka-console-producer.sh --broker-list kafka:9092 --topic my-topic  --producer.config=$HOME/client.properties
```

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mstruk: The idea is that 90% of the time people will configure at most 3 reporters. That way there will be no need to reallocate the underlying array to accommodate more. But then this is not a performance critical code and if you think it distracts from reading the code we can remove it. In all practical sense it has zero impact on performance.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mstruk: I don't know how you decide on the importance of the particular configuration option. In this case the warning is most welcome because it is almost certainly an omission that STRIMZI_METRIC_REPORTERS was not configured. With metric enabled and not exported to JMX there is no way to collect by any mechanism (as fas as I know).

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mstruk: In this case we can as stringVal cannot be null here.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mstruk: No, because I still want to maintain java 1.8 compatibility which is what Kafka still supports as far as I know.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/Services.java
mstruk: Maybe one day it will be. There are at the moment no threads running in there so I simply release the memory. You think this should be called differently or the docs should be different or you think the implementation should be different?

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mimaison: If you think this is meaningful optimization, I'd add a comment justifying why 3 was picked. When I saw it, it looked like a random number to me.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mimaison: If not setting it generates warnings, it looks pretty important to me. That said I'm not sure what conventions are used in other parts of the Strimzi code so if none of the maintainers pick on this, maybe we can ignore it.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mimaison: Can this run client side? I though Strimzi was now developed and tested with Java 17.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
scholzj: I would agree with Mickael that this seems a bit weird. I guess when you put it that enabling metrics but not configuring the metrics reporter would be wrong, that makes sense. But why does the user need to:
* Enable the metrics
* And configure the metrics reporter

Why not do just one and have the other done automatically by the Oauth library? I guess I'm missing some obvious reason why it is not possible, but it would be good to understand it.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mstruk: This can run client side. It can also run in a Kafka distribution other than Strimzi Kafka images.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mstruk: For OAuth metrics I didn't want them to be enabled by default in order to avoid potential issues (performance or memory consumption for example). And that did allow us to very easily work around CruiseControl issue - everything was fine until someone enabled OAuth metrics. The solution was - don't enable OAuth metrics.

OAuth metrics also has finer control on metrics collection. You can enable metrics globally, or for just particular listeners, or just for KeycloakAuthorizer.

Kafka metrics are enabled by default and which of them are enabled is controlled by [metrics.recording.level](https://kafka.apache.org/documentation/#producerconfigs_metrics.recording.level). But as far as I understand there is no way to completely turn them off.

For OAuth we could change the default for the configuration (enabled by default and you can disable it), but if we otherwise keep  backwards compatibility around `metric.reporters` this will result in things crashing / acting up by default when non-compatible metric reporters (such as CruiseControl metric reporter) are configured, as we would double-instantiate them.
We could then enable metrics gathering only when some reporters are configured via 'strimzi.oauth.metric.reporters'. However, if we made this change then absence of `strimzi.oauth.metric.reporters` if we want to maintain backwards compatibility with `metrics.reporters` would mean that we have to instantiate `JmxReporter` each and every time even if `metric.reporters` is not set.

I would prefer for now to NOT have OAuth metrics enabled by default. Let's have it opt-in until we're confident that there are no issues when using them (like the issue this PR is addressing of things crashing when a reporter not compatible with OAuth metrics is configured).

scholzj: I can understand why OAuth metrics should not be enabled by default. But why does the user need to configure the `strimzi.oauth.metric.reporters`? Isn't it Strimzi OAuth who consumes this option anyway? So why not set it automatically.

mstruk: Ah, yes that's a great point. I was motivated by aligning with where [Kafka's handling of `metric.reporters`](https://cwiki.apache.org/confluence/display/KAFKA/KIP-830%3A+Allow+disabling+JMX+Reporter) is headed.
With Kafka 4.0 there will be no exporting of (automatically enabled) metrics to JMX unless a specific reporter is configured that provides such an export.

We can opt to not go this way to avoid needing to set two configurations (Kafka does not have that problem since its metrics are always enabled). We could for example handle omission of `strimzi.oauth.metric.reporters` by automatically creating JmxReporter. If user specifies `strimzi.oauth.metric.reporters` we could only instantiate the configured reporters, but we would not check if JmxReporter is part of the config, and we would not automatically instantiate it if not present in the config (that's different from current `metric.reporters` handling when we automatically instantiate JmxReporter if it is not part of the `metric.reporters` list).

That probably makes more sense in our case as then it is not necessary to always set it, and user should consult the documentation anyway to determine the behaviour and that it is not the same as for Kafka metrics configured via `metric.reporters`.
Also the presence of `strimzi.oauth.metric.reporters=` (configured and set to empty value) could then mean to NOT instantiate a JmxReporter if for some reason one really wants to have no reporters attached to the OAuth Metrics object (and thus no export to JMX).

testsuite/common/src/main/java/io/strimzi/testsuite/oauth/common/TestUtil.java
mstruk: You're absolutely right. That's a dumb code. I'll remove the redundant prepareRegex() and use matches()

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mstruk: In fact the allocated object is never added to. A remnant of refactoring ... I'll remove it.

oauth-common/src/main/java/io/strimzi/kafka/oauth/services/OAuthMetrics.java
mimaison: Also keep in mind that hopefully that should all disappear in the near future once KIP-877 gets in Kafka. So maybe we want to keep things simple.

mstruk: Ok, I'll modify the logic as described above. It means no backwards compatibility in the sense that `metric.reporters` will be ignored. But if `strimzi.oauth.metric.reporters` is set it will be used, if not set, `JmxReporter` will automatically be instantiated.

README.md
mstruk: Added examples to README.

README.md
```diff
@@ -1225,21 +1225,73 @@ Configuring the metrics
 
 By default, the gathering and exporting of metrics is disabled. Metrics are available to get an insight into the performance and failures during token validation, authorization operations and client authentication to the authorization server. You can also monitor the authorization server requests by background services such as refreshing of JWKS keys and refreshing of grants when `KeycloakAuthorizer` is used.
 
-You can enable metrics for token validation on the Kafka broker or for client authentication on the client by setting the following JAAS option to `true`:
+You can enable metrics for token validation, and keycloak authorization on the Kafka broker or for client authentication on the client by setting the following JAAS option to `true`:
```
scholzj, line 1228: Should it be _Keycloak_? Or should it be marked as a code to indicate some configuration option?

README.md
```diff
@@ -1225,21 +1225,73 @@ Configuring the metrics
...
 - `oauth.enable.metrics` (e.g.: "true")
 
-You can enable metrics for `KeycloakAuthorizer` by setting an analogous option in Kafka broker's `server.properties` file:
+You can also enable metrics only for keycloak authorization by setting an analogous option in Kafka broker's `server.properties` file:
```
scholzj, line 1231: Same as above ...

scholzj (approved):

README.md
mstruk: I changed it to: `KeycloakAuthorizer`

README.md
mstruk: I changed it to: `KeycloakAuthorizer`

mstruk:
@tombentley @mimaison Any more unaddressed issues here?

README.md
```diff
@@ -1225,21 +1225,73 @@ Configuring the metrics
...
 
-The OAuth metrics also honor the `metric.reporters`, `metrics.num.samples`, `metrics.recording.level` and `metrics.sample.window.ms` configurations of Kafka runtimes. When OAuth metrics are enabled the OAuth layer creates duplicate instances of `JmxReporter` and the configured `MetricReporter`s since at the moment there is no other way to integrate with the existing metrics system of Kafka runtimes. 
+The configuration option `strimzi.oauth.metric.reporters` on the Kafka broker has to be configured as an env variable or a system property. Using it on the Kafka broker inside `server.properties` does not work reliably due to multiple pluggability mechanisms that can be used (authorizer, authentication callback handler, inter-broker client).
+Some of these mechanisms get the `server.properties` filtered so only configuration recognised by Kafka makes it through. However, this is a global OAuth Metrics configuration, and it is initialized on first use by any of the components, using the configuration provided to that component.
```
tombentley, line 1244: > Some of these mechanisms get the `server.properties` filtered so only configuration recognised by Kafka makes it through.

Not a comment about this PR, really an observation about Kafka being inconsistent/the `Configurable` contract being rather weak. If `Configurable` itself isn't going to specify whether `configure()` is getting the `originals`, or some filtered subset then it should be documented on each subclass what is passed. Thoughts @mimaison?

tombentley (approved):
LGTM, thanks Marko.

README.md
mimaison: I thought the "contract" was to always pass `originals` to plugins so they can have custom configurations.
https://issues.apache.org/jira/browse/KAFKA-7588

tombentley: That @mstruk which plugin did you find where the `originals` weren't propagated to `configure()`?

mstruk: It is the inter-broker communication client that gets initialized with the already parsed Configuration object, which then no longer contains Map<String, String> but Map<String, ?> and the unrecognised config options are missing.

PR #194, opened 2023-06-28
Principal extraction from nested username claim was broken

mstruk:
When extracting a user id from JWT token by using `oauth.username.claim` or `oauth.fallback.username.claim` it only worked for top level attributes, not for nested attributes. For example, by configuring: `"oauth.username.claim=auth.userid"`, and given a JWT token:
```
{
    ...
   "auth": {
      "userid": "alice"
   }
}
```
Extraction would not find 'userid' key under top level 'auth' object, rather it was looking for 'auth.userid' top level key.

This PR adds an option to use JsonPath to target nested attributes.
If the claim specification starts with an opening square bracket '[', it is interpreted as a JsonPath query.
Otherwise, it is interpreted as a top level attribute name.

```
userId                    ... use top level attribute named 'userId'
user.id                   ... use top level attribute named 'user.id'
$userid                   ... use top level attribute named '$userid'
['userInfo'].id           ... use nested attribute 'id' under 'userInfo' top level attribute
['user.info']['user.id']  ... use nested attribute 'user.id' under 'user.info' top level attribute
['user.info'].['user.id'] ... use nested attribute 'user.id' under 'user.info' top level attribute (optional dot)
```

### DIFF
--- a/oauth-common/src/main/java/io/strimzi/kafka/oauth/common/PrincipalExtractor.java
+++ b/oauth-common/src/main/java/io/strimzi/kafka/oauth/common/PrincipalExtractor.java
@@ -10,7 +10,7 @@ import static io.strimzi.kafka.oauth.common.JSONUtil.getClaimFromJWT;
 
 /**
  * An object with logic for extracting a principal name (i.e. a user id) from a JWT token.
- *
+ * <p>
  * First a claim configured as <code>usernameClaim</code> is looked up.
  * If not found the claim configured as <code>fallbackUsernameClaim</code> is looked up. If that one is found and if
  * the <code>fallbackUsernamePrefix</code> is configured prefix the found value with the prefix, otherwise not.
@@ -29,7 +29,7 @@ public class PrincipalExtractor {
     /**
      * Create a new instance
      *
-     * @param usernameClaim Attribute name for an attribute containing the user id to lookup first
+     * @param usernameClaim Attribute name for an attribute containing the user id to lookup first. Use '.' to refer to nested child objects.
      * @param fallbackUsernameClaim Attribute name for an attribute containg the user id to lookup as a fallback
      * @param fallbackUsernamePrefix A prefix to prepend to the value of the fallback attribute value if set
      */
@@ -49,13 +49,13 @@ public class PrincipalExtractor {
         String result;
 
         if (usernameClaim != null) {
-            result = getClaimFromJWT(json, usernameClaim);
+            result = getClaimFromJWT(usernameClaim, json);
             if (result != null) {
                 return result;
             }
 
             if (fallbackUsernameClaim != null) {
-                result = getClaimFromJWT(json, fallbackUsernameClaim);
+                result = getClaimFromJWT(fallbackUsernameClaim, json);
                 if (result != null) {
                     return fallbackUsernamePrefix == null ? result : fallbackUsernamePrefix + result;
                 }
--- a/oauth-common/src/main/java/io/strimzi/kafka/oauth/common/PrincipalExtractor.java
+++ b/oauth-common/src/main/java/io/strimzi/kafka/oauth/common/PrincipalExtractor.java
@@ -106,6 +106,26 @@ public class PrincipalExtractor {
         return usernameClaim != null || fallbackUsernameClaim != null || fallbackUsernamePrefix != null;
     }
 
+    /**
+     * This method parses the configured principal extraction specification using a syntax similar to JSONPath.
+     * A claim (an attribute inside a JSON object) is targeted in arbitrary levels of nested depth.
+     * For specifying nested claims the specification has to start with an opening square bracket and end with a closing square bracket.
+     * The segments are separated by a dot '.'  (e.g.: [topClaim].[subClaim]).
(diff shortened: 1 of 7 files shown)


scholzj (changes requested):
I think this deserves some description to expain what this change does. You seem to have really just swicthed two parameters around without any change to the method you call. So it is hard to understand the logic behind it.

mstruk:
@scholzj Thanks for pointing this out. Upon a second look I realised that the fix is not good as it potentially introduces backwards compatibility issues. I'll describe the problem in more detail, and I'm working on a proper fix.

scholzj (approved):

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/PrincipalExtractor.java
```diff
@@ -88,4 +105,60 @@ public String toString() {
...
+            parsed.add(removeQuotesAndSpaces(spec.substring(pos, epos)));
+            pos = epos + 1;
+            epos = skipSpaces(spec, pos);
+            if (epos >= spec.length()) {
+                return parsed;
+            }
+            if (spec.charAt(epos) != '.') {
+                throw new IllegalArgumentException("Failed to parse usename claim spec: '" + spec + "' (Missing '.' at position: " + epos + ")");
+            }
+            pos = spec.indexOf("[", epos + 1);
```
tombentley, line 137: How does this handle the input `[foo].bar[baz]`?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/PrincipalExtractor.java
```diff
@@ -88,4 +105,60 @@ public String toString() {
...
+        LinkedList<String> parsed = new LinkedList<>();
+        int pos = 0;
+        int epos = 0;
+        while (pos != -1) {
+            pos += 1;
+            epos = spec.indexOf("]", pos);
+            if (epos == -1) {
+                throw new IllegalArgumentException("Failed to parse username claim spec: '" + spec + "' (Missing ']')");
+            }
+            parsed.add(removeQuotesAndSpaces(spec.substring(pos, epos)));
```
tombentley, line 128: How does this handle the input `['foo'bar']`?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/PrincipalExtractor.java
mstruk: This should be illegal. It should throw an exception. I'll fix it. Once you start your claim spec with '[' the characters between ']' and '[' should be limited to arbitrary number of spaces and a single '.'.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/PrincipalExtractor.java
```diff
@@ -88,4 +105,60 @@ public String toString() {
     public boolean isConfigured() {
         return usernameClaim != null || fallbackUsernameClaim != null || fallbackUsernamePrefix != null;
     }
+
+    private static List<String> parseClaimSpec(String spec) {
```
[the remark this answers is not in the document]
mstruk: Added javadoc to the method.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/PrincipalExtractor.java
mstruk: Fixed.

mstruk:
@tombentley I hope I adequately addressed the comments. WDYT?

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/PrincipalExtractor.java
tombentley: I accept that this is an edge case, but I find it really surprising anything should accept `'foo'bar'` as if it were a literal containing the characters `foo'bar`. I can't think of any examples (e.g. in programming languages) where things are parsed where such input wouldn't be rejected. Such parsing is usually based on tokenisation and the idea of paired delimiters. If people need to include `'` in the middle of a literal like that the right way to do it is require escaping. You're violating a very very common expectation here. That would be OK if there was a valid use for supporting this, but you agree that no one should want or need this flexibility. 

I think it's far safer to reject inputs like this, since it's more than likely due to a user messing up their config than something intentional. And rejection would appear to be easy enough: Anything inside the quotes should match `[a-zA-Z0-9.]+` or similar.

In general it's much easier to be strict from that start, and relax once its clear that the rules are too strict for valid use cases that users have, than to try to restrict inputs later on. By that point you don't know what weird and wonderful things people might be using and expecting to continue working.

mstruk: @tombentley I modified the implementation to use JsonPath for nested attributes. This should remove any inconsistencies in special character matching ...

tombentley (approved):
Thanks @mstruk

PR #199, opened 2023-07-11
Try to reset JAVA_HOME to fix broken s390x and ppc64le builds

mstruk:


mstruk:
Looks like travis-ci.com people fixed things at their end.

PR #201, opened 2023-07-27
Support disabling the Accept header when requesting Json Web Key Sets.

hmadison:
For certain servers which provide Json Web Key Sets, such as the Kubernetes API Server, if the `Accept` header sent by the client is present it is expected to be `application/jwk-set+json`. Based off of [the discussion](https://cloud-native.slack.com/archives/CMH3Q3SNP/p1690308865023079) in the Strimzi slack channel, I've added an option which allows for users to skip sending the `Accept` header when making the JWKS request. 

I've also updated the test documentation to cover how to use MiniKube and Helm to simplify the testing process and included an example of how to use this feature to authenticate into a Kafka cluster with a Service Account's token. To validate this locally, you can apply `kafka-oauth-single-authz-service-accounts.yaml` to a Kubernetes cluster with the operator installed and using a version of `strimzi-kafka-oauth` installed then:
- Generate a JWT for the default service account in the default namespace with `kubectl create token default`.
- Enter the Kafka pod via `kubectl exec -it pod/test-cluster-kafka-0 -- /bin/bash`.
- Connect to the cluster via one of the included scripts using`env OAUTH_ACCESS_TOKEN="<token>" bin/kafka-console-producer.sh --broker-list test-cluster-kafka-bootstrap.default:9092 --topic my-topic --producer-property 'security.protocol=SASL_PLAINTEXT' --producer-property 'sasl.mechanism=OAUTHBEARER' --producer-property 'sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required;' --producer-property 'sasl.login.callback.handler.class=io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler'`


examples/kubernetes/README.md
```diff
@@ -35,6 +35,12 @@ They assume Keycloak is used as an authorization server, with properly configure
   A single node Kafka cluster with OAuth 2 authentication with OAuth metrics enabled.
   See [README-metrics.md]() for how to setup this example.
 
+* `kafka-oauth-single-authz-service-accounts.yaml`
+
+  A single node Kafka cluster using [service account tokens](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#service-account-tokens) for authorization and the `simple` authorizer.
+  It requires that the `kube-root-ca.crt` be copied from its ConfigMap to a Secret:
+    
+      kubectl get configmap/kube-root-ca.crt -o=json | jq -r '.data."ca.crt"' | kubectl create secret generic kube-root-ca --from-file=ca.crt=/dev/stdin
```
scholzj, line 43: Is this really something what works with every single Kubernetes cluster based on Service accounts only? Or does it really on OpenID provider?

examples/kubernetes/README.md
```diff
@@ -35,6 +35,12 @@ They assume Keycloak is used as an authorization server, with properly configure
   A single node Kafka cluster with OAuth 2 authentication with OAuth metrics enabled.
   See [README-metrics.md]() for how to setup this example.
 
+* `kafka-oauth-single-authz-service-accounts.yaml`
+
+  A single node Kafka cluster using [service account tokens](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#service-account-tokens) for authorization and the `simple` authorizer.
```
scholzj, line 40:
```suggestion
  A single node Kafka cluster using [service account tokens](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#service-account-tokens) for authentication and the `simple` authorizer.
```

examples/kubernetes/kafka-oauth-single-authz-service-accounts.yaml
```diff
@@ -0,0 +1,78 @@
...
+    A role binding which allows for anonymous access to those endpoints
+    from consuming applications.
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: oidc-reader
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: User
+    name: "system:anonymous"
```
scholzj, line 29: This should have the service account of the Kafka brokers I guess and not anonymous?

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/ServerConfig.java
```diff
@@ -83,6 +83,11 @@ public class ServerConfig extends Config {
      */
     public static final String OAUTH_GROUPS_CLAIM_DELIMITER = "oauth.groups.claim.delimiter";
 
+    /**
+     * "oauth.include.accept.header"
+     */
+    public static final String OAUTH_INCLUDE_ACCEPT_HEADER = "oauth.include.accept.header";
```
scholzj, line 89: Should there be some documentation for this option?

mstruk: This is just a string constant in `ServerConfig.java`. Following the existing established pattern the option should be documented in `JaasServerOauthValidatorCallbackHandler.java` and in `README.md`.
But actually, I see this as a global option which should be moved from `ServerConfig.java` to `Config.java`, and taken into account also in `KeycloakRBACAuthorizer.java`, and in `JaasClientOauthLoginCallbackHandler`.

mstruk (commented):
Thanks for the PR. As I mentioned in another comment I think this new option constant should really live in `Config.java`, and functionally be used not only in `JWTSignatureValidator.java` but also in `OAuthIntrospectionValidator.java`, and in `KeycloakRBACAuthorizer.java` and `JaasClientOauthLoginCallbackHandler`.

In `KeycloakRBACAuthorizer.java` the analogous option should also be introduced `strimzi.authorization.include.accept.header` that falls back to `oauth.include.accept.header` the same way some other options in there do.

examples/kubernetes/README.md
hmadison: The Kubernetes api-server supports [OpenID Connect](https://kubernetes.io/docs/concepts/security/service-accounts/#authenticating-credentials) to validate the credentials which (since version 1.22) get projected into a pod based off of it's service account. You don't need anything outside of a Kubernetes cluster to make use of this authentication.

examples/kubernetes/kafka-oauth-single-authz-service-accounts.yaml
hmadison: Since `strimizi-kafka-oauth` isn't setting its Bearer token to the one projected into the container, you need to allow `system:anonymous` to access the endpoints needed to retrieve the OpenID Connect settings and the Json Web Key Sets.

hmadison:
@mstruk I've made the requested changes. Do you have any idea why the s390x portion of the build is currently breaking?

examples/kubernetes/kafka-oauth-single-authz-service-accounts.yaml
scholzj: If that is not enabled for everyone by default, it means this is likely a security risk? Should it really be used here? Shouldn't you instead use the broker's service account?

scholzj: I guess the same applies to `system:serviceaccount:default:default` -> maybe you should use some better example SA then the default one.

I'm also not sure I follow the relation ship of this example to the PR here. How do they relate? You do not seem to use any new flag in the example to disable the `Accept` header. So while the Kubernetes example is interesting, I do not follow the relationship here.

examples/kubernetes/README.md
scholzj: Ok, interesting. I will have to try that.

examples/kubernetes/kafka-oauth-single-authz-service-accounts.yaml
hmadison: I'm setting the value of `oauth.include.accept.header` to `false` via `spec.kafka.jvmOptions.javaSystemProperties` on line 63.

From my perspective, the ODIC and JWKS endpdoints in a "normal" application are exposed over the public internet. If someone is in a position to make a web request from inside of the kubernetes cluster to these endpoints, they are also going to be able to access the token for the service account and make those requests with authentication. If we want to use the brokers service account, then we should back this example out of this pull request and I'll open a new one which includes a dedicated authenticator for Kubernetes service accounts.

mstruk (commented):
Some minor issues, otherwise it looks good to me.

examples/kubernetes/kafka-oauth-single-authz-service-accounts.yaml
```diff
@@ -0,0 +1,78 @@
...
+rules:
+  - nonResourceURLs: ["/.well-known/openid-configuration", "/openid/v1/jwks"]
+    verbs: ["get", "post"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: oidc-reader-binding
+  annotations:
+  kubernetes.io/description: |
```
mstruk, line 19: Still not ok. It's this line (19) that should be indented in order to be a child of `annotations`. Currently it is a child of `metadata` which is incorrect.

examples/kubernetes/kafka-oauth-single-authz-service-accounts.yaml
mstruk: We'll need to add the `includeAcceptHeader` config option to `oauth` listener authentication and `keycloak` authorization when we integrate the next OAuth release into the Strimzi Operator (another PR). We can then improve the example Kafka CR.

scholzj: > I'm setting the value of oauth.include.accept.header to false via spec.kafka.jvmOptions.javaSystemProperties on line 63.

I missed that part. As mentioned by Marko, this will need to be added tot he options after the next OAuth release. Until then it is a bit misleading since it will not work with OAuth 0.13.

> From my perspective, the ODIC and JWKS endpdoints in a "normal" application are exposed over the public internet. If someone is in a position to make a web request from inside of the kubernetes cluster to these endpoints, they are also going to be able to access the token for the service account and make those requests with authentication. If we want to use the brokers service account, then we should back this example out of this pull request and I'll open a new one which includes a dedicated authenticator for Kubernetes service accounts.

It seems that from Kubernetes perspective it is not something what they thing should be public. Also, thing like this will discourage users from using it as it does not _look secure_. If this works on any Kubernetes cluster, I think this would be a great usecase. But I think we need to work out the details to make sure it uses the service account and does not connect over as anonymous. So my vote would be to move this to a separate PR and do this through the broker service account.

examples/kubernetes/kafka-oauth-single-authz-service-accounts.yaml
hmadison: Based off of @scholzj's feedback, I've backed this example out and will open up a subsequent PR with a better "out of the box" experience around service account usage.

examples/kubernetes/kafka-oauth-single-authz-service-accounts.yaml
hmadison: That's fine by me. I'll back this change out and start on a dedicated Kubernetes service account configuration which supports using the pod's service account token to authenticate with the api server.

examples/kubernetes/README.md
```diff
@@ -35,6 +35,12 @@ They assume Keycloak is used as an authorization server, with properly configure
   A single node Kafka cluster with OAuth 2 authentication with OAuth metrics enabled.
   See [README-metrics.md]() for how to setup this example.
 
+* `kafka-oauth-single-authz-service-accounts.yaml`
+
+  A single node Kafka cluster using [service account tokens](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#service-account-tokens) for authentication and the `simple` authorizer.
+  It requires that the `kube-root-ca.crt` be copied from its ConfigMap to a Secret:
+    
+      kubectl get configmap/kube-root-ca.crt -o=json | jq -r '.data."ca.crt"' | kubectl create secret generic kube-root-ca --from-file=ca.crt=/dev/stdin
```
scholzj, line 43: I guess we should remove this part as well for the time being?

scholzj (commented):
One nit. LGTM otherwise.

scholzj (approved):

mstruk:
I added some tests to the testsuite and fixed some additional issues I found while writing tests: https://github.com/hmadison/strimzi-kafka-oauth/pull/1

mstruk (approved):

mstruk:
@hmadison Thank you very much for the contribution!

PR #211, opened 2023-10-30
Oauth client assertion support

robbertvanwaveren:
This PR adds support for Kafka clients to obtain an access token by authenticating to the authorization server with client assertion as specified by https://www.rfc-editor.org/rfc/rfc7523 and https://www.rfc-editor.org/rfc/rfc7521. The assertion can be provided by an external mechanism and available as a file on the file system or it can be explicitly set through OAuth configuration before running the Kafka client.

In addition, the static access token or refresh token can now be provided as a file on the local filesystem.

The following new configuration options have been added by this PR:
- `oauth.client.assertion`
- `oauth.client.assertion.location`
- `oauth.client.assertion.type`
- `oauth.refresh.token.location`
- `oauth.access.token.location`

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,9 +147,7 @@ This is true for configuring the server side of the Kafka Broker, as well as for
 The authentication configuration specific to the Strimzi Kafka OAuth can be specified as part of JAAS configuration in the form of JAAS parameter values. 
 The authorization configuration for `KeycloakAuthorizer` is specified as `server.properties` key-value pairs.
 Both authentication and authorization configuration specific to Strimzi Kafka OAuth can also be set as ENV vars, or as Java system properties.
-The limitation here is that authentication configuration specified in this manner can not be listener-scoped. 
-
-Note that property-values starting with `env:` are interpreted as references to existing ENV vars.
+The limitation here is that authentication configuration specified in this manner can not be listener-scoped.
 
 ### Configuring the Kafka Broker authentication
 
--- a/README.md
+++ b/README.md
@@ -149,6 +149,8 @@ The authorization configuration for `KeycloakAuthorizer` is specified as `server
 Both authentication and authorization configuration specific to Strimzi Kafka OAuth can also be set as ENV vars, or as Java system properties.
 The limitation here is that authentication configuration specified in this manner can not be listener-scoped. 
 
+Note that property-values starting with `env:` are interpreted as references to existing ENV vars.
+
 ### Configuring the Kafka Broker authentication
 
 Note: Strimzi Kafka OAuth can not be used for Kafka Broker to Zookeeper authentication. It only supports Kafka Client to Kafka Broker authentication (including inter-broker communication).
@@ -926,19 +928,46 @@ Strimzi Kafka OAuth supports four ways to configure authentication on the client
 
 #### Client Credentials
 
-The first is to specify the client ID and secret configured on the authorization server specifically for the individual client deployment. This is also called `client credentials grant`.
+The first is to specify Client Credentials. This requires that a client is configured on the authorization server specifically for the individual client deployment. This is also called `client credentials grant`.
 
 This is achieved by specifying the following:
 - `oauth.client.id` (e.g.: "my-client")
+
+together with one of authentication options below 
+
+When client starts to establish the connection with the Kafka Broker it will first obtain an access token from the configured Token Endpoint, authenticating with the configured client ID and configured authentication option using client_credentials grant type.
+
+##### Option 1: Using a Client Secret 
+
+Specify the client secret.
+
 - `oauth.client.secret` (e.g.: "my-client-secret")
 
-When client starts to establish the connection with the Kafka Broker it will first obtain an access token from the configured Token Endpoint, authenticating with the configured client ID and secret using client_credentials grant type.
+##### Option 2: Using a Client Assertion (a.k.a. private_key_jwt)
+
+Specify the client assertion (JWT token) either directly through
+
+- `oauth.client.assertion`
+
+or pointing to a file on the filesystem
+
+- `oauth.client.assertion.location`
+
+the exact type of the token must also be communicated to the token endpoint and defaults to `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`. 
+
+This can be overridden using property
+
+-  `oauth.client.assertion.type` (i.e. use `urn:ietf:params:oauth:client-assertion-type:saml2-bearer` for SAML 2 tokens)
 
 #### Refresh Token
 
-The second way is to manually obtain and set a refresh token:
+The second way is to manually obtain and set a refresh token either directly through
 
 - `oauth.refresh.token`
+
+or pointing to a file on the filesystem
+
+- `oauth.refresh.token.location`
 
 When using this approach you are not limited to OAuth2 client_credentials grant type for obtaining a token.
 You can use a password grant type and authenticate as an individual user, rather than a client application.
@@ -948,9 +977,13 @@ When client starts to establish the connection with the Kafka Broker it will fir
 
 #### Access Token
 
-The third way is to manually obtain and set an access token:
+The third way is to manually obtain and set an access token either directly through:
 
 - `oauth.access.token`
+
+or pointing to a file on the filesystem
+
+- `oauth.access.token.location`
 
 Access tokens are supposed to be short-lived in order to prevent unauthorized access if the token leaks.
 It is up to you, your environment, and how you plan to run your Kafka client application to consider if using long-lived access tokens is appropriate.
--- a/README.md
+++ b/README.md
@@ -953,7 +953,9 @@ or pointing to a file on the filesystem
 
 - `oauth.client.assertion.location`
 
-the exact type of the token must also be communicated to the token endpoint and defaults to `urn:ietf:params:oauth:client-assertion-type:jwt-bearer` (which is specified in RFC-7523). 
+The file contains secrets in plain text and should have proper permissions set - not readable by others.
+
+The exact type of the token must also be communicated to the token endpoint and defaults to `urn:ietf:params:oauth:client-assertion-type:jwt-bearer` (which is specified in RFC-7523). 
 
 This can be overridden using property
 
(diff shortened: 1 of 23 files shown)


robbertvanwaveren:
this PR now has DCO signoff and grants maintainer edit rights

scholzj (commented):
Can we have the description and the PR name reflect what it does?

mstruk:
I added a better description.

scholzj (approved):

scholzj:
@mstruk Same as the other PR. The Travis tests do not seem to run here.

README.md
```diff
@@ -926,20 +928,47 @@ Strimzi Kafka OAuth supports four ways to configure authentication on the client
...
+
+Specify the client assertion (JWT token) either directly through
+
+- `oauth.client.assertion`
+
+or pointing to a file on the filesystem
+
+- `oauth.client.assertion.location`
+
+the exact type of the token must also be communicated to the token endpoint and defaults to `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`.
```
tombentley, line 956:
```suggestion
the exact type of the token must also be communicated to the token endpoint and defaults to `urn:ietf:params:oauth:client-assertion-type:jwt-bearer` (which is specified in RFC-7523). 
```

README.md
```diff
@@ -926,20 +928,47 @@ Strimzi Kafka OAuth supports four ways to configure authentication on the client
 
 #### Client Credentials
 
-The first is to specify the client ID and secret configured on the authorization server specifically for the individual client deployment. This is also called `client credentials grant`.
+The first is to specify Client Credentials. This requires that a client is configured on the authorization server specifically for the individual client deployment. This is also called `client credentials grant`.
 
 This is achieved by specifying the following:
 - `oauth.client.id` (e.g.: "my-client")
+
+together with one of authentication options below
```
tombentley, line 936:
```suggestion
together with one of authentication options below.
```

README.md
```diff
@@ -926,20 +928,47 @@ Strimzi Kafka OAuth supports four ways to configure authentication on the client
...
+
+or pointing to a file on the filesystem
+
+- `oauth.client.assertion.location`
+
+the exact type of the token must also be communicated to the token endpoint and defaults to `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`. 
+
+This can be overridden using property
+
+-  `oauth.client.assertion.type` (i.e. use `urn:ietf:params:oauth:client-assertion-type:saml2-bearer` for SAML 2 tokens)
```
tombentley, line 960:
```suggestion
-  `oauth.client.assertion.type` (i.e. use `urn:ietf:params:oauth:client-assertion-type:saml2-bearer`, specified in RFC-7522, for SAML 2 tokens)
```

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
```diff
@@ -193,11 +215,20 @@ private long getHttpRetryPauseMillis(ClientConfig config, int retries) {
         return retryPauseMillis;
     }
 
+    private TokenProvider createProvider(final String token, final String tokenLocation) {
+        if (tokenLocation != null) {
```
tombentley, line 219: Do we want to accept the case where both parameters are non-null? It's ambiguous what the user's intent is in this case.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
```diff
@@ -167,6 +167,90 @@ public static TokenInfo loginWithClientSecret(URI tokenEndpointUrl, SSLSocketFac
...
+     * @param principalExtractor A PrincipalExtractor to use to determine the principal (user id)
+     * @param scope A scope to request when authenticating
+     * @param audience An 'audience' attribute to set on the request when authenticating
+     * @param includeAcceptHeader Should we skip sending the Accept header when making outbound http requests
+     * @return A TokenInfo with access token and information extracted from it
+     * @throws IOException If the request to the authorization server has failed
+     * @throws IllegalStateException If the response from the authorization server could not be handled
+     */
+    @SuppressWarnings("checkstyle:ParameterNumber")
+    public static TokenInfo loginWithClientAssertion(URI tokenEndpointUrl, SSLSocketFactory socketFactory,
```
tombentley, line 191: Can we put each parameter on a separate line.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
```diff
@@ -167,6 +167,90 @@ public static TokenInfo loginWithClientSecret(URI tokenEndpointUrl, SSLSocketFac
...
+     * @param metrics A MetricsHandler object to receive metrics collection callbacks
+     * @param retries A maximum number of retries if the request fails due to network, or unexpected response status
+     * @param retryPauseMillis A pause between consecutive requests
+     * @param includeAcceptHeader Should we skip sending the Accept header when making outbound http requests
+     * @return A TokenInfo with access token and information extracted from it
+     * @throws IOException If the request to the authorization server has failed
+     * @throws IllegalStateException If the response from the authorization server could not be handled
+     */
+    @SuppressWarnings("checkstyle:ParameterNumber")
+    public static TokenInfo loginWithClientAssertion(URI tokenEndpointUrl, SSLSocketFactory socketFactory,
```
tombentley, line 225: Can we put each parameter on a separate line.

tombentley (commented):
We shouldn't assume that people trying to use this understand what they're doing. Even if they do it's easy to become focussed on "making it work" and forget about all the steps needed to actually make it secure. so I think:

1. the README should recommend the use of filesystem permissions to protect tokens on disk.
2. we should consider, when the OS is unix, checking the file permissions and failing if they token files are readable by all, or the group. Obviously this doesn't provide any protection on non-unix, but it's better than nothing in my opinion.

mstruk:
> 1. the README should recommend the use of filesystem permissions to protect tokens on disk.

Good point, I'll add some docs to address this.

> 2. we should consider, when the OS is unix, checking the file permissions and failing if they token files are readable by all, or the group. Obviously this doesn't provide any protection on non-unix, but it's better than nothing in my opinion.

Failing is a harsh reaction to this. Especially if an external mechanism takes care of mounting a filesystem so it's all automatic and possibly out of your control. I would prefer to log a warning, but not fail. WDYT?

tombentley:
> Failing is a harsh reaction to this. Especially if an external mechanism takes care of mounting a filesystem so it's all automatic and possibly out of your control. I would prefer to log a warning, but not fail. WDYT?

The 'possibly out of your control point' is a good one. So I guess logging a warning is reasonable.

mstruk:
The 'env:ENV_VARIABLE_NAME' feature doesn't seem to add any capability that couldn't be performed before. Also rather than conflating such a feature with the current PR it is better to introduce it separately if at all.

Also, there is a possibility of unexpected behaviour, for example using `oauth.client.secret=\"env:Hfs67vfv.M8a\"` would automatically trigger a lookup of `Hfs67vfv.M8a` env variable which in this case was not the intention.

It is already the case that when a configuration option is resolved the resolution tries several different override possibilities in a specific order as implemented [here](https://github.com/strimzi/strimzi-kafka-oauth/blob/0.14.0/oauth-common/src/main/java/io/strimzi/kafka/oauth/common/Config.java#L143-L190) and documented [here](https://github.com/strimzi/strimzi-kafka-oauth/tree/0.14.0#configuring-the-oauth2). 

First, if the system property exists with the same name it overrides the JAAS configured attribute. Second, if the env variable with the same name exists that is used. Third, if the env variable exists with the name equal to configuration key converted to all caps and underscores then that one is used. Only if there are no overrides via System properties and env vars the config parameter as configured via JAAS is used.

According to these rules using `oauth.client.assertion.location="env:TOKEN_FILE"` can be achieved by setting the following env variable in the client's execution environment: `OAUTH_CLIENT_ASSERTION_LOCATION="$TOKEN_FILE"` rather than set as a JAAS parameter.

robbertvanwaveren:
I don’t see how the logic you refer to performs the substitution of the TOKEN_FILE env variable for its actual content.

The original intent of this PR is support for MS Azure Workload Identity which injects certain env vars into the execution environment. Which is the main (only) practic use-case for this PR.

Note that this means that these env var names are not configurable so these should be piped towards this libs configs parama/vars.

mstruk:
@robbertvanwaveren Can you give an example or provide a failing test?

Substitution happens in your kafka-client.yaml (or whatever you call the yaml file that you use) when setting environment vars for the pod.
Or in your bash startup script used as an entry point for your client.

oauth-client/src/main/java/io/strimzi/kafka/oauth/client/JaasClientOauthLoginCallbackHandler.java
mstruk: I aded a logic that if both are set, the location (file) is used rather than the token, and warning is logged.

oauth-common/src/main/java/io/strimzi/kafka/oauth/common/OAuthAuthenticator.java
mstruk: I aded logic that if both are set, the location (file) is used rather than the token, and warning is logged.

mstruk:
I'm thinking of renaming the `oauth.*.location` config options to `oauth.*.path`. The reasoning is that `location` is too generic, it can mean a file on the local filesystem or a url or something completely different as one of the values from a set of supported values. By saying `path` it makes it super clear (at least to me) that it is the path to the file on the local filesystem.

I wonder what others think?

mstruk:
Actually that won't work as there are existing previous options that refer to files and are called `oauth.*.location` (e.g. `oauth.ssl.truststore.location`) and I prefer to keep consistency in the naming.

mstruk:
@tombentley Do you think this needs another review (specifically, the file permissions check)?

PR #213, opened 2023-11-15
Update testsuite/README.md

mstruk:
Remove references to Arquillian which has been replaced with `testcontainers` a long time ago.


mstruk:
Addressed by #220

PR #221, opened 2024-02-02
Update Keycloak used in testsuite and examples to latest (23.0.5) 

mstruk:


mstruk:
> Thanks @mstruk - I will get this further investigated at our end.

@rposts There are more details about the things I tried at #222.

scholzj (approved):

rposts:
> Thanks @mstruk - I will get this further investigated at our end.

@mstruk could we use release binaries as is? This worked for me:
```
git clone -b 23.0.5 https://github.com/keycloak/keycloak
cd keycloak/container/
docker build . -t quay.io/keycloak/keycloak:23.0.5
docker run --rm -it quay.io/keycloak/keycloak:23.0.5 -V
Keycloak 23.0.5
JVM: 17.0.10 (Red Hat, Inc. OpenJDK 64-Bit Server VM 17.0.10+7-LTS)
OS: Linux 5.4.0-170-generic s390x
```

mstruk:
@rposts Sure, if that works for you. I didn't find Keycloak image builds for `s390x` in quay.io so I assumed this still needs to be built manually for your purposes. If you find this solution to work, it's fine with me to do it this way.

PR #222, opened 2024-02-05
Troubleshoot s390x testsuite build

mstruk:


rposts:
@mstruk @mstruk - Just wondering if we can directly use `keycloak` binaries instead of building from scratch on `s390x`? This seems to work for me:
```
git clone -b 23.0.5 https://github.com/keycloak/keycloak
cd keycloak/container/
docker build . -t quay.io/keycloak/keycloak:23.0.5
docker run --rm -it quay.io/keycloak/keycloak:23.0.5 -V
Keycloak 23.0.5
JVM: 17.0.10 (Red Hat, Inc. OpenJDK 64-Bit Server VM 17.0.10+7-LTS)
OS: Linux 5.4.0-170-generic s390x
```

PR #231, opened 2024-02-29
Add client support for SASL extensions

mstruk:
Add support for passing SASL extensions via OAuth configuration options, by using a prefix: `oauth.sasl.extension.`

### DIFF
--- a/testsuite/mockoauth-tests/src/test/resources/simplelogger.properties
+++ b/testsuite/mockoauth-tests/src/test/resources/simplelogger.properties
@@ -2,5 +2,6 @@
 org.slf4j.simpleLogger.logFile=target/test.log
 org.slf4j.simpleLogger.showDateTime=true
 org.slf4j.simpleLogger.dateTimeFormat=yyyy-MM-dd HH:mm:ss:SSS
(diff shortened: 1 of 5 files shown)


testsuite/mockoauth-tests/src/test/resources/simplelogger.properties
```diff
@@ -2,5 +2,6 @@
 org.slf4j.simpleLogger.logFile=target/test.log
 org.slf4j.simpleLogger.showDateTime=true
 org.slf4j.simpleLogger.dateTimeFormat=yyyy-MM-dd HH:mm:ss:SSS
+org.slf4j.simpleLogger.log.org.apache.kafka.common.security=TRACE
```
scholzj, line 5: Is this intentional change?

scholzj (approved):

testsuite/mockoauth-tests/src/test/resources/simplelogger.properties
mstruk: It is indeed. The new test makes sure the SASL extensions are in fact sent by the SASL client layer by inspecting the log.

PR #239, opened 2024-05-07
Update README.md

### LINKED ISSUE #238: Check path `kcadm.sh` in keycloak
https://github.com/strimzi/strimzi-kafka-oauth/blob/229daee85b096804d16e2904c8c0f1add599cc99/examples/README.md?plain=1#L25

Not sure where the difference comes from (maybe change that came with version upgrade of the imagequay.io/keycloak/keycloak), but for me it was this dir that contained the `kcadm.sh`:
```
cd /opt/keycloak/
```

So maybe change  `cd /opt/jboss/keycloak` to `cd /opt/keycloak/`?

flo-kn:
suggestion to fix path to sh script

closes #238

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -22,7 +22,7 @@ Assuming `keycloak` container is up and running you can do the following.
 Connect to 'keycloak' container to get access to `kcadm.sh` tool:
 
     docker exec -ti docker-keycloak-1 /bin/sh
-    cd /opt/jboss/keycloak
+    cd /opt/keycloak/
 
 Set the server endpoint url:
 


scholzj (commented):
This likely relates to the container image that is used. So I do not think you can change it just like that without providing some more details.

flo-kn:
It's dependent on the keycloak image, or image version. Wouldn't it make most sense to have the path to the script similar to what ever image is used here:

https://github.com/strimzi/strimzi-kafka-oauth/blob/229daee85b096804d16e2904c8c0f1add599cc99/examples/docker/keycloak/compose.yml#L6

..to serve consistency with rest of the example? That was my assumption here, requesting that change.

scholzj:
If that is the path in the images used in the examples then yes, that would make sense. But you have to properly explain it so that others understand it and not just change the path without any proper explanation here or in the issue.

mstruk:
The proposed fix looks good. When Keycloak container version was bumped last time (from pre-quarkus version of Keycloak) this line should have been adjusted as well.

A deeper question for @flo-kn - Do you find value in this README.md file? It seems to me to unnecessarily rely on Keycloak command line client in order to demonstrate how Keycloak generates different Payload section of JWT token using the predefine realm (clients and users already configured) and depending on the client / user used.

`kcadm.sh` tool here is used purely as a readily available OAuth client. The same can be achieved using `curl` / `wget` and would be more useful as these two are readily available in application container images that give access to shell.

flo-kn:
To your question @mstruk Generally, I think it is good to have some lines in the  README.md explaining how to verify functionality of the installation or in this case parts of the installation. Think if it's `curl`, `wget`, or `kcadm.sh` on the container instance depends on personal taste.

If you'd ask me for  a suggestion to  improve the readme a biet: In this case, I think setting another heading would be beneficial to give it more structure so that user knows in advance what next lines will be about.  E.g. around this area would be a suitable place: https://github.com/strimzi/strimzi-kafka-oauth/blob/229daee85b096804d16e2904c8c0f1add599cc99/examples/README.md?plain=1#L20

## Verify Keycloak Functionality
Before we start with the demos the let's check first that keycloak is working correctly. 

To do that connect to 'keycloak' contai....

flo-kn:
> If that is the path in the images used in the examples then yes, that would make sense. But you have to properly explain it so that others understand it and not just change the path without any proper explanation here or in the issue.

@scholzj  Just updated the issue description. Think adding more explanation to the guide itself other then the correct directory does not bring any additional benefit.

scholzj:
Discussed on the community call: @mstruk will take the ideas and use them as improvements to the README files. This PR should be closed.

PR #250, opened 2024-11-20
Fix Keycloak urls in example to work with recent versions of Keycloak

mstruk:
The OAuth2 endpoints in Keycloak examples still use the urls as they were by default on legacy versions of Keycloak (e.g. 19.0.3-legacy). In the testsuite we moved some time ago to start using recent versions of Keycloak (e.g. 23.0.x) and we adjusted the OAuth2 endpoint urls there to use the default structure compatible with recent Keycloak releases.

Specifically, where before the url would look like: `https://localhost:8443/auth/realms/demo`
it now looks like: `https://localhost:8443/realms/demo`


testsuite/docker/kafka/config/log4j.properties
```diff
@@ -91,5 +91,5 @@ log4j.logger.state.change.logger=TRACE, stateChangeAppender
 log4j.additivity.state.change.logger=false
 
 # Access denials are logged at INFO level, change to DEBUG to also log allowed accesses
-log4j.logger.kafka.authorizer.logger=INFO, authorizerAppender
+log4j.logger.kafka.authorizer.logger=DEBUG, authorizerAppender
```
scholzj, line 94: Intentional?

scholzj (commented):
I think this deserves a better description. For example expain why did the URL changed? (new Keycloak version I guess?)

testsuite/docker/kafka/config/log4j.properties
mstruk: I guess we can wait and see if CI fails, and if it does revert this change.

mstruk:
Travis is failing for `s390x` only, and I've seen it fail this way for #249 as well, so I'm pretty sure it's not related to the changes in this PR. And actually, the DEBUG logging adds very minimal extra logs - only some additional info that comes handy when debugging authorization, so I think we can leave it as is.

scholzj (approved):

PR #257, opened 2025-01-29
Improve kerberos tests in testsuite to work with Podman

mstruk:


scholzj (approved):
LGTM as long as it works.

PR #258, opened 2025-01-29
Update default testsuite kafka image to 0.45.0-kafka-3.9.0

mstruk:


mstruk:
Travis tests are passing except for s390x where the execution is so long that it times out. I think we can count that as passing.

scholzj (approved):

PR #259, opened 2025-02-11
Bump dependencies versions in examples and testsuite, and some cleanup

mstruk:


ppatierno (approved):

mstruk:
Unclear why the test on `amd64` with `openjdk17` is failing. The exact error is:

[ERROR] Plugin org.springframework.boot:spring-boot-maven-plugin:3.3.7 or one of its dependencies could not be resolved: The following artifacts could not be resolved: org.springframework.boot:spring-boot-maven-plugin:jar:3.3.7 (absent): org.springframework.boot:spring-boot-maven-plugin:jar:3.3.7 was not found in https://maven-central.storage-download.googleapis.com/maven2/ during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of google-maven-central has elapsed or updates are forced -> [Help 1]

But the artifact was found in maven repository for all other java / platform combinations. It seems like a corrupt local maven repository cache.

ppatierno (approved):

PR #265, opened 2025-04-11
Bump json-smart version to 2.5.2

mstruk:
`json-smart` is being pulled in transitively by JsonPath project and flagged by CVE scanners for CVE-2024-57699 warning. Even though we don't think we are affected due to how we use JsonPath, the fact that scanners flag the dependency makes users nervous.

This is a temporary override, to be removed once a new version of JsonPath is released.
See: https://github.com/json-path/JsonPath/pull/1030


ppatierno (approved):
LGTM

PR #280, opened 2025-09-03
Update dependencies - libraries with CVEs, image versions used in examples and testsuite

mstruk:


ppatierno (approved):
LGTM, assuming it works ;-)

PR #283, opened 2025-10-22
Fix for `"this.authorizerMetrics" is null` issue in Kafka 4.1.0

mstruk:
This PR fixes the issue when using `KeycloakAuthorizer` in Kafka 4.1.0 when delegation to `StandardAuthorizer` is enabled (`strimzi.authorization.delegate.to.kafka.acl=true`).

Whenever the authorization decision is delegated to `StandardAuthorizer` the following exception happens:

```
java.lang.NullPointerException: Cannot invoke "org.apache.kafka.metadata.authorizer.StandardAuthorizer$AuthorizerMetrics.recordAuthorizerMetrics(org.apache.kafka.server.authorizer.AuthorizationResult)" because "this.authorizerMetrics" is null
	at org.apache.kafka.metadata.authorizer.StandardAuthorizer.authorize(StandardAuthorizer.java:153) ~[kafka-metadata-4.1.0.jar:?]
	at io.strimzi.kafka.oauth.server.authorizer.KeycloakRBACAuthorizer.delegateIfRequested(KeycloakRBACAuthorizer.java:469) ~[kafka-oauth-keycloak-authorizer-0.17.0.jar:?]
	at io.strimzi.kafka.oauth.server.authorizer.KeycloakRBACAuthorizer.authorize(KeycloakRBACAuthorizer.java:340) ~[kafka-oauth-keycloak-authorizer-0.17.0.jar:?]
	at io.strimzi.kafka.oauth.server.authorizer.KeycloakAuthorizer.authorize(KeycloakAuthorizer.java:227) ~[kafka-oauth-keycloak-authorizer-0.17.0.jar:?]
...
```

Kafka 4.1.0 added new metrics pluggability mechanism and updated `StandardAuthorizer` to implement `Monitorable` interface and expect a `PluginMetrics` object to be provided by instantiating server via `Monitorable.withPluginMetrics(PluginMetrics pluginMetrics)`.

For `KeycloakAuthorizer` to be deployable to Kafka 4.0.0 where `Monitorable` interface is not present, and to also properly work on Kafka 4.1.0 where `Monitorable.withPluginMetrics()` is expected to be called on `StandardAuthorizer` the only way to address this without having multiple versions of `KeycloakAuthorizer` is to use reflection.

It is not possible for `KeycloakAuthorizer` to implement `Monitorable` and still be compatible with Kafka 4.0.0, thus the `PluginMetrics` instance that we set on `StandardAuthorizer` when running in Kafka 4.1.0 is not passed in and backed by the `Metrics` object of `BrokerServer` or `ControllerServer` but by a separate internally managed `Metrics` object used by all the Strimzi OAuth pluggable classes.

### DIFF
--- a/oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java
+++ b/oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java
@@ -106,7 +106,7 @@ public class KeycloakAuthorizer implements ClusterMetadataAuthorizer {
     private void initPluginMetricsIfNeeded() {
         Class<?> monitorableClass = null;
         try {
-            // Check if StandardAuthorizer implements Monitorable, which can only be true if running on Kafka 4.1.0+
+            // Check if Monitorable is present. This indicates if we're running on Kafka 4.1.0+
             monitorableClass = Class.forName("org.apache.kafka.common.metrics.Monitorable");
         } catch (Exception e) {
             log.debug("Monitorable class not present. PluginMetrics initialisation skipped on StandardAuthorizer");
@@ -119,7 +119,7 @@ public class KeycloakAuthorizer implements ClusterMetadataAuthorizer {
                 pluginMetrics = method.invoke(null, delegate, metrics, "authorizer.class.name", "role", determineRole());
             }
         } catch (Throwable e) {
-            log.warn("Failed to initialise PluginMetrics on StandardAuthorizer", e);
+            throw new ConfigException("Failed to initialise PluginMetrics on StandardAuthorizer (consider setting `strimzi.authorization.delegate.to.kafka.acl=false`): ", e);
         }
     }
 


oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java
```diff
@@ -85,13 +89,74 @@ public void configure(Map<String, ?> configs) {
...
+    /**
+     * Initialise and set PluginMetrics on StandardAuthorizer if it implements Monitorable
+     * Use of reflection is needed since the classes are only available starting with Kafka 4.1.0
+     * Once this class implements Monitorable the use of reflection will no longer be necessary, and the passed-in
+     * PluginMetrics object will be passed on to StandardAuthorizer.withPluginMetrics().
+     */
+    private void initPluginMetricsIfNeeded() {
+        Class<?> monitorableClass = null;
+        try {
+            // Check if StandardAuthorizer implements Monitorable, which can only be true if running on Kafka 4.1.0+
```
mimaison: The comment is a bit misleading. This block checks if the Monitorable class is present, ie, we're running >=4.1

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java
```diff
@@ -85,13 +89,74 @@ public void configure(Map<String, ?> configs) {
...
+        }
+        try {
+            if (monitorableClass != null && monitorableClass.isAssignableFrom(delegate.getClass())) {
+                Class<?> pluginClass = Class.forName("org.apache.kafka.common.internals.Plugin");
+                Metrics metrics = Services.getInstance().getMetrics().getKafkaMetrics();
+                Method method = pluginClass.getMethod("wrapInstance", Object.class, Metrics.class, String.class, String.class, String.class);
+                pluginMetrics = method.invoke(null, delegate, metrics, "authorizer.class.name", "role", determineRole());
+            }
+        } catch (Throwable e) {
+            log.warn("Failed to initialise PluginMetrics on StandardAuthorizer", e);
```
mimaison: Isn't it fatal if we reach this point?

testsuite/docker/hydra-import/scripts/start.sh
```diff
@@ -23,7 +23,7 @@ cat < /hydra/certs/ca.crt >> /etc/ssl/certs/ca-certificates.crt
 
 URI=${HYDRA_URI}
 if [ "" == "${URI}" ]; then
-    URI="http://${HYDRA_HOST:-hydra}:{SERVE_ADMIN_PORT:-4445}/clients"
```
mimaison: It seems many changes in the files below are not directly related to Kafka 4.1?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java
mimaison: What about `// Check if Monitorable is present. This indicates if we're running on Kafka 4.1.0+`?

mstruk:
> How much effort would it take to stop using non-public Kafka APIs?

Can you be more specific here? We wrap `StandardAuthorizer` and take control of its lifecycle. `StandardAuthorizer` is not considered public API based on its package. But we need a way to delegate to it and control its lifecycle. So what are even the options?

oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java
mstruk: Modified so it fails fast.

PR #286, opened 2025-10-23
Use kafka 4.1.0 image in testsuite and minor other fixes

mstruk:


mstruk:
The tests are supposed to fail, and the failure is exactly the one expected - due to '"this.authorizerMetrics" is null' issues.

Issue #281, opened 2025-10-01
Add s390x build status in README


sangitanalkar:
Hello, 

As per proposal mentioned here ([link](https://github.com/strimzi/proposals/blob/main/099-drop-travis-ci-and-testing-for-ppc-and-s390x.md)) , s390x architecture is `supported` in Azure pipelines but `not tested`.
We’ve set up a CI job for building and testing strimzi-kafka-oauth on s390x via OSU infrastructure: [IBMZ CI Job](https://ibmz-ci.osuosl.org/job/Strimzi_Kafka_Oauth_IBMZ_CI/). The job runs daily and is actively monitored.
We are looking forward to add s390x build badge to README.

Would you be open to adding a webhook in the strimzi-kafka-oauth GitHub repo to trigger this Jenkins job on merges?
Also, I’d be happy to raise a PR to add a s390x build badge to the README. Please let us know your thoughts.

Thanks.

PR #287, opened 2025-10-29
Add s390x CI build status badge in README.md

sangitanalkar:
Raising PR to address issue https://github.com/strimzi/strimzi-kafka-oauth/issues/281

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Strimzi Kafka OAuth modules provide support for OAuth2 as authentication mechani
   - [Using the metrics with Prometheus](#using-the-metrics-with-prometheus)
   - [Some examples of PromQL queries](#some-examples-of-promql-queries)
 - [Demo](#demo)
+- [Community Testing for Linux on IBM Z s390x architecture](#community-testing-for-linux-on-ibm-z-s390x-architecture)
   
 <!-- /TOC -->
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,12 @@ Strimzi OAuth for Apache Kafka
 [Apache Kafka®](https://kafka.apache.org) comes with basic OAuth2 support in the form of SASL based authentication module which provides client-server retrieval, exchange and validation of access token used as credentials.
 For real world usage, extensions have to be provided in the form of JAAS callback handlers which is what Strimzi Kafka OAuth does.
 
+### Architecture support
+#### Linux on IBM Z (s390x) :
```
scholzj: * This basically suggests that the OAuth library supports only s390x, which is not the case. So we would need to rework the title and come up with something clearer. Or add the information about the other _supported_ architectures as well. But given this is pure Java, I think speaking about supported architectures does not much sense.
* Also, the way you added this essentially nests the sentence on line 19 under the `Linux on IBM Z (s390x)` chapter. That makes no sense. You would need to add it below it, have another title, or move this to its own section somewhere around the Building and Installing chapters, where it would make more sense I guess.

sangitanalkar:
> @sangitanalkar we discussed this on the community call and would like to use "Community Testing" rather than "Architecture Support" so could you update the PR with that change. See [strimzi/strimzi-kafka-operator#12044 (comment)](https://github.com/strimzi/strimzi-kafka-operator/pull/12044#issuecomment-3468813718)

Hi @katheris I have updated as per above suggestion.

PR #296, opened 2026-02-26
Rewrite testsuite to follow Java's testing best practices

Frawless:
**Current state & Motivation**
Current testsuite uses `docker compose` to spinup the testing environment. It handles Kafka, Keycloak, Hydra, and MockOAuth components completely via docker. This makes managing changes not easy to do, because we cannot do it per-test in Java and we basically influence all tests tests with every change.

The next problem is requirements on host level. To be able to run the tests properly we need to do changes in `/etc/hosts` which is not user friendly at all.

Another problem is with tests structure. Currently the tests are in very strange nested structure. There is usually one major test class that contains single test (method annotated with `@Test` annotation) and this single method calls other methods that either calls another real test scenarios or do the test scenarios directly. The result is very hard-to-read scenarios that are potentially influenced by each other, because some simply rely on previous "test". 

Overall this structure makes the following actions quite hard:
- Understand existing scenarios
- Add new scenarios
- Debug failures locally
- Have isolated env for each scenario

**Scope of changes in this PR**
This PR contains mainly the following major changes:
- Migrate from `junit4` to `junit6`
- Create a way how to easily create isolated environment for test classes/tests with `test-containers`
- Migrate from docker-compose to Java test-container for auth servers and to Strimzi test-container for Kafka
- Migrate from Kafka with around 14 listeners to Kafka with just one listener (limitation of Strimzi test-container)
- Add annotations that setup testing env easily for the users
- Improve logging of the tests cases (logs from containers are stored for every test case, test log from junit runner is stored next to the containers as well, failed tests are highlighted in the logs structure, use proper logger instead of System.println)
- Removed required changes to /etc/hosts for users and also in our CI

**What is not part of this PR**
- Migrate running tests against different Kafka versions from maven config to Java
- Unit-test changes except migration to newer junit version

I understand this PR is huge and the review won't be trivial, it was hard for me as well. I open it as a draft to get first feedback from SMEs (@mstruk from oauth side and @see-quick from test-container side). 

**Any review and feedback is welcome!**


mimaison:
I'm not a Strimzi maintainer but a PR this size is just not practical to review.

I just took a very quick glance and, at least at first sight, it seems it could be broken up into smaller PRs. I think that a task like updating to junit6 could be its own PR. And this could be the case for some of the other "phases" and bullet points you listed in the description.

ppatierno:
Yeah, I agree with @mimaison . This seems to be a pretty too big PR which is kind of impossible to review (at least it would be for me). I think we should avoid something like this in general unless it's not possible to break it in several PRs.

mstruk:
@Frawless Thank you for this PR. I am reviewing it in its current form. It is big chunk to bite and it is taking a while for me to grasp all the changes, but I can see it was a good effort and a lot of work to get all the details in place.

Some general observations ... It seems that all the tests have been rewritten to the new approach so it seems that there is no lack-of-functionality as a possible reason for not migrating to this approach.

On a first note I observed the execution times almost doubled which may be a result of each test getting running individually, and also the produced logs are excessively verbose so I think we should reduce logging somewhat.

This is also a response to:
> There is usually one major test class that contains single test (method annotated with `@Test` annotation) and this single method calls other methods that either calls another real test scenarios or do the test scenarios directly.

Current tests are structured the way they are to minimize the testsuite execution time, and logs are minimized to reduce log size. In the past the CI integration for PRs had issues with both execution time and log size so it was necessary to do something about it. I guess these factors might still play a role but we may be operating within more generous quotas?

In terms of getting this PR merged in a more managable fashion, ideally we would structure PRs in such a way that we first provide general foundation and then for each migrated test we provide an individual PR. That would make it easier to catch any inadvertent functional changes in the tests which is something we absolutely want to ensure. It is not enough that the new testsuite passes, it also has to keep every single test functionally equivalent.

In fact the very act of reviewing this PR requires me to pretty much do this same thing locally step-by-step so one option is that I create such PRs as I make incremental changes locally as I perform the review.

Frawless:
> On a first note I observed the execution times almost doubled which may be a result of each test getting running individually, and also the produced logs are excessively verbose so I think we should reduce logging somewhat.
> Current tests are structured the way they are to minimize the testsuite execution time, and logs are minimized to reduce log size. In the past the CI integration for PRs had issues with both execution time and log size so it was necessary to do something about it. I guess these factors might still play a role but we may be operating within more generous quotas?

Currently we have kinda unlimited execution time. Of course if we will run things 24/7 it will be a problem at some point, but running pipeline 30 minutes instead of 15 is completely fine. Also we can utilize junit groups and run multiple jobs to speed-up verification, but I don't this is necessary atm.

> In terms of getting this PR merged in a more managable fashion, ideally we would structure PRs in such a way that we first provide general foundation and then for each migrated test we provide an individual PR. That would make it easier to catch any inadvertent functional changes in the tests which is something we absolutely want to ensure. It is not enough that the new testsuite passes, it also has to keep every single test functionally equivalent.
> 
> In fact the very act of reviewing this PR requires me to pretty much do this same thing locally step-by-step so one option is that I create such PRs as I make incremental changes locally as I perform the review.

In case you are willing to do that it would be great! I don't have much bandwith now to go through it, but I should be able to do reviews. We can follow same pattern as @im-konge did with few Strimzi and Bridge stuff (one branch, open PRs against and then merge one huge PR with reviewed parts). I think some parts can be easily cherry-picked as I tried to do some kind of phases, but overall it didn't make sense to me to do just a half of the work without clearly see if it makes sense or not. 

Guess we can close this one and use it as a reference starting point for refactor.

Frawless:
@mstruk btw this https://github.com/strimzi/test-container/issues/192 should make some things easier to do as TC should allow multiple listeners once implemented so the cases where multiple listeners were used can be migrated 1:1 as well. I think in this PR I changed the scenarios a little so there might be some differences in functionality coverage, but during implementation I thought it is fine.

mstruk:
> Guess we can close this one and use it as a reference starting point for refactor.

Let's keep it open in this Draft stage for now, and close it after my PRs are created.

see-quick:
> > Guess we can close this one and use it as a reference starting point for refactor.
> 
> Let's keep it open in this Draft stage for now, and close it after my PRs are created.

Also let me know if having support for multiple listeners would help you (or if it's needed) and if we should consider add support for it in the test container.

PR #314, opened 2026-05-08
Add Pluggable JWT Key Type Support and add OKP (Ed25519) Implementation

mstruk:
This PR introduces a pluggable architecture for JWT signature validation, enabling extensible support for different key types through Java ServiceLoader. The implementation adds optional support for OKP (Octet Key Pair) keys, specifically Ed25519, while refactoring existing support for RSA and EC keys such that they are loaded automatically and maintain backwards compatibility.

### Core Architecture Refactoring
- Introduced SigningKeyProvider SPI for extensible JWT key type support via Java ServiceLoader
- Added JCASigningKeyProvider as the built-in provider handling RSA and EC keys (in oauth-common)
- Refactored JWTSignatureValidator to use the pluggable provider architecture instead of hardcoded key type handling

### New Optional Module: oauth-okp-support
- Created oauth-okp-support module with Ed25519 (OKP) key support using Google Tink library
- Implemented OKPSigningKeyProvider for OKP key validation
- Enabled via Maven profile `-Pokp-support` to keep the dependency optional

### Comprehensive Testing
- Added end-to-end Keycloak authentication tests with OKP keys (OKPTest)
- Added mock OAuth tests for OKP validation (OKPValidationTest)
- Created demo-okp-realm.json Keycloak realm configuration for testing
- Added OKPTestGroup test category for selective test execution

### CI/CD Integration
- Added dedicated CI lane for OKP-enabled builds to ensure both configurations are tested:
  - Default build without OKP support
  - Feature-enabled build with OKP support running OKP-specific tests

### Benefits
- Extensibility: New JWT key algorithms can be added as separate opt-in modules without modifying core code
- Backward Compatibility: Existing RSA and EC key support remains unchanged in the core library
- Optional Dependencies: OKP support and its Google Tink dependency are only included when explicitly enabled
- Future-Proof: Architecture supports adding additional key types or different implementations of existing ones as needed


.github/actions/build/test-oauth/action.yml
```diff
@@ -55,9 +61,20 @@ runs:
...
+    - name: Build OKP-enabled testsuite dependencies
+      if: ${{ inputs.okpSupport == 'true' }}
+      shell: bash
+      run: mvn -e -V -B install -f testsuite -pl common,mock-oauth-server,docker -am -Pokp-support --no-transfer-progress
+
+    - name: Run mockoauth tests with OKP support
+      if: ${{ inputs.okpSupport == 'true' }}
+      shell: bash
+      run: mvn -e -V -B install -f testsuite -pl mockoauth-tests,keycloak-auth-tests -Pokp-support -Dgroups=io.strimzi.testsuite.oauth.common.OKPTestGroup --no-transfer-progress
+
```
Frawless, line 77: This looks to me as quite messy solution. Can't we simply use just different profile and let maven handle dependencies, etc. internally? Or add it into make?

Frawless (dismissed):
Also I would prefer to finish https://github.com/strimzi/strimzi-kafka-oauth/pull/296 or at least some parts of it (junit5/6 migration, test-container usage). You are adding here more and more conflicts and new scenarios that are now written in format that probably only you fully understand.

.github/actions/build/test-oauth/action.yml
mstruk: What we could do is to always only run the full testsuite with `-Pokp-support` profile. But that requires the `-Pokp-support` profile to always be used while building the project in order to also make the `oauth-okp-support.jar` available. For users building by themselves they would still be able to build the project and the testsuite with or without the profile, but for CI purposes we would always have `okp-support` profile enabled. Which may be what we want in the end - to also automatically publish the `oauth-okp-support` artifact as part of the release.

With such a solution we can then also remove the OKPTestGroup which is there as an optimization to avoid full double run of the testsuite.

It would then just be:
```
- name: Build & Test Java
      shell: bash
      run: mvn -e -V -B install -DskipTests -Pokp-support
      env:
        BUILD_REASON: ${{ github.event_name }}
        BRANCH: ${{ github.ref }}
        TESTCONTAINERS_RYUK_DISABLED: "TRUE"
        TESTCONTAINERS_CHECKS_DISABLE: "TRUE"
...

- name: Run Testsuite
      shell: bash
      run: mvn -e -V -B install -f testsuite --no-transfer-progress -Pokp-support

```
without the OKP specific steps.

mstruk:
> Also I would prefer to finish https://github.com/strimzi/strimzi-kafka-oauth/pull/296 or at least some parts of it (junit5/6 migration, test-container usage). You are adding here more and more conflicts and new scenarios that are now written in format that probably only you fully understand.

We can put this PR on hold until #296 is merged and then rebase, refactor the tests part.

.github/actions/build/test-oauth/action.yml
Frawless: Well the group can be specified in profile to skip unrelated tests. So run the build with specific profile and then tests should be fine. Proper docs probably resolve the issue that users build oauth with expectation of okp support will be there by default, but without specifing the profile.

Frawless:
> > Also I would prefer to finish #296 or at least some parts of it (junit5/6 migration, test-container usage). You are adding here more and more conflicts and new scenarios that are now written in format that probably only you fully understand.
> 
> We can put this PR on hold until #296 is merged and then rebase, refactor the tests part.

I think it is reasonable. But if there is a strong need or different opinions on that, we can talk about it on next community call for example or on slack.

mstruk:
One remaining dilemma about the implementation approach in this feature is the following. Currently the implementation uses `ServiceLoader` pattern to scan for `SigningKeyProviders` on the classpath. It finds `JCASigningKeyProvider` coming from the `kafka-oauth-common` jar (which is always present). Then if `kafka-oauth-okp-support` jar is present it also finds the `OKPSigningKeyProvider`.

There is a potential unpredictability here in that multiple providers for the same signing key type are possible and their loading order is determined by classpath order of jars. In such a case this may result in a different provider being registered than the one expected by the developer. 

Existing Kafka plugins generally always have a specific configuration property where plugins should be listed, and they are loaded and registered in the order they were listed.

In the same vein an implementation more in-line with this Kafka configuration pattern would use another configuration property to control ordering e.g.:

````
oauth.jwks.key.providers=io.strimzi.kafka.oauth.validator.JCASigningKeyProvider,io.strimzi.kafka.oauth.validator.okp.OKPSigningKeyProvider
````

I think in practice that would only be needed if wanting to replace the default `JCASigningKeyProvider` with a different implementation and with current implementation one would have to make sure that the new jar would be on the classpath before `kafka-oauth-common` jar in order for the replacement implementation to take precedence.

Any opinions on this?

PR #315, opened 2026-05-21
[Feature]support configurable introspection token param

### LINKED ISSUE #299: Support configurable introspection token parameter name (for providers requiring access_token)
## Summary

When using `strimzi-kafka-oauth` with introspection/tokeninfo, the request body currently uses the `token` parameter name.  
This works for many RFC7662-compatible providers, but some providers require `access_token` instead.

In these environments, broker authentication fails even though the bearer token is valid.

---

## Environment / Use Case

We are integrating Strimzi Kafka with CI OAuth (IdBroker-style tokeninfo endpoint).  
Current setup uses broker-side OAuth introspection (`strimzi-kafka-oauth`).

Our tokeninfo endpoint expects form data:

`access_token=<token>`

But current introspection call from Strimzi uses:

`token=<token>`

---

## Observed Behavior

- Introspection endpoint returns `400 Bad Request`
- Error message is similar to: `Missing access token.`
- Broker reports runtime token validation failure
- Kafka clients receive `SaslAuthenticationException`

---

## Expected Behavior

There should be a way to configure the introspection request token parameter name, so Strimzi can interoperate with providers that require `access_token` instead of `token`.

---

## Proposed Design

Add an optional config property:

`oauth.introspection.token.param.name`

### Rules

- Default value: `token` (preserve backward compatibility)
- If configured (for example `access_token`), use that key in introspection POST form body
- If empty/invalid, safely fall back to `token`
- Scope only introspection request body field name
- No change to JWT local validation, JWKS, issuer/audience checks, etc.

### Example

```properties
oauth.introspection.endpoint.uri=https://idbroker.example.com/oauth2/v1/tokeninfo
oauth.introspection.token.param.name=access_token

### PR DESCRIPTION
**Pull request by litiliu**
This PR addresses an interoperability gap in introspection-based validation.

Problem solved:
The broker always sent introspection requests as token=<access_token>.
Some OAuth/tokeninfo providers require a different parameter name (for example access_token).
Result: introspection call failed (HTTP 400 / missing access token), so authentication could not proceed.
this close https://github.com/strimzi/strimzi-kafka-oauth/issues/299

What was changed:
Added a new server config: `oauth.introspection.token.param.name`.
Kept backward compatibility by defaulting to token when the option is not set.
Wired the new option through server config parsing and validator initialization.
Updated OAuthIntrospectionValidator to build the POST body using the configured parameter name instead of a hardcoded token.
Included the new parameter in IntrospectionValidatorKey (equals/hash/config id hash) to prevent incorrect validator/cache reuse across different configs.
Updated tests and README to cover/document the new behavior.

Impact:
Enables Strimzi OAuth to work with providers that expect non-default introspection parameter names (e.g., access_token) without changing client behavior.
Existing deployments remain unaffected unless they opt in to the new setting.

### DIFF
--- a/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
+++ b/oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
@@ -52,9 +52,11 @@ import static io.strimzi.kafka.oauth.validator.TokenValidationException.Status;
 public class OAuthIntrospectionValidator implements TokenValidator {
 
     private static final Logger log = LoggerFactory.getLogger(OAuthIntrospectionValidator.class);
+    private static final String DEFAULT_INTROSPECTION_TOKEN_PARAM_NAME = "token";
 
     private final String validatorId;
     private final URI introspectionURI;
+    private final String introspectionTokenParamName;
     private final String validIssuerURI;
     private final URI userInfoURI;
     private final String validTokenType;
@@ -90,6 +92,7 @@ public class OAuthIntrospectionValidator implements TokenValidator {
      * @param clientSecret The secret of the OAuth2 client representing this Kafka broker - used to authenticate to the introspection endpoint using Basic authentication
      * @param bearerTokenProvider The provider of the bearer token as an alternative to clientId and secret of the OAuth2 client representing this Kafka broker - used to authenticate to the introspection endpoint using Bearer authentication
      * @param introspectionEndpointUri The introspection endpoint url at the authorization server
+     * @param introspectionTokenParamName The request parameter name used to send token to introspection endpoint
      * @param socketFactory The optional SSL socket factory to use when establishing the connection to authorization server
      * @param verifier The optional hostname verifier used to validate the TLS certificate by the authorization server
      * @param principalExtractor The object used to extract the username from the attributes in the server's response
@@ -113,6 +116,7 @@ public class OAuthIntrospectionValidator implements TokenValidator {
                                        String clientSecret,
                                        TokenProvider bearerTokenProvider,
                                        String introspectionEndpointUri,
+                                       String introspectionTokenParamName,
                                        SSLSocketFactory socketFactory,
                                        HostnameVerifier verifier,
                                        PrincipalExtractor principalExtractor,
@@ -133,6 +137,7 @@ public class OAuthIntrospectionValidator implements TokenValidator {
         this.validatorId = checkValidatorId(id);
 
         this.introspectionURI = checkIntrospectionUri(introspectionEndpointUri);
+        this.introspectionTokenParamName = parseIntrospectionTokenParamName(introspectionTokenParamName);
 
         this.socketFactory = checkSocketFactory(socketFactory);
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java [hunk 1]
```diff
@@ -296,7 +313,7 @@ public TokenInfo validate(String token) {
 
         String authorization = generateAuthorizationHeader();
 
-        StringBuilder body = new StringBuilder("token=").append(token);
+        StringBuilder body = new StringBuilder(introspectionTokenParamName).append("=").append(token);
```
mstruk: Let's urlencode `introspectionTokenParamName` here because it's provided by the user configuration. Actually let's encode the `token` as well as it can also be provided by the user configuration.

oauth-common/src/main/java/io/strimzi/kafka/oauth/validator/OAuthIntrospectionValidator.java
```diff
@@ -260,6 +266,17 @@ private String checkValidatorId(String validatorId) {
         return validatorId;
     }
 
+    private String parseIntrospectionTokenParamName(String tokenParamName) {
```
mstruk: For the sake of not repeating yourself I would make this method a more generic utility method that can be put in IOUtil:
```
public static String trimmedNonEmptyValueOrDefault(String value, String defaultValue) {
   if (value == null) {
      return defaultValue;
   }
   value = value.trim();
   return value.isEmpty() ? defaultValue : value;
}
```

Then invoke it as:

    trimmedNonEmptyValueOrDefault(introspectionTokenParamName, DEFAULT_INTROSPECTION_TOKEN_PARAM_NAME)

And also reuse it in [JaasServerOauthValidatorCallbackHandler.java‎](https://github.com/strimzi/strimzi-kafka-oauth/pull/315/changes#diff-0af30cbf06d7deee32bafafb11afc8d0f5ac5242878a6e7746f4f315ca94ea11)

oauth-server/src/main/java/io/strimzi/kafka/oauth/server/JaasServerOauthValidatorCallbackHandler.java
```diff
@@ -436,6 +442,14 @@ private String setupIntrospectionValidator(
         return effectiveConfigId;
     }
 
+    private String normalizeIntrospectionTokenParamName(String introspectionTokenParamName) {
```
mstruk: Remove this method and instead use:

    trimmedNonEmptyValueOrDefault(config.getValue(ServerConfig.OAUTH_INTROSPECTION_TOKEN_PARAM_NAME), DEFAULT_INTROSPECTION_TOKEN_PARAM_NAME)

mstruk (changes requested):
Generally it looks good, but there are some changes needed.

mstruk (approved):
The changes look good. Thank you for the contribution @litiliu

im-konge (approved):
The code changes LGTM, thanks :)

scholzj:
Should we merge this or focus first on the test fixes? I added this to the community call agenda.

litiliu:
Thanks for adding it to the agenda. From my side this PR is ready and includes tests for the new behavior. If the test fixes refer to any failure caused by this PR, please point me to the failing test/log and I’ll take a look. @scholzj

scholzj:
@litiliu Apologies for confusing you ... the point was mainly about this PR https://github.com/strimzi/strimzi-kafka-oauth/pull/296 which does several fixes to the tests n general. Not strictly to any tests in this PR. Sorry.

scholzj:
Discussed on the community call on 11.6.2026: This can be merged. Thanks for the PR.

